Issue #5, opened 2017-03-30
Add the `request-id` option to all plugins to enable tracking plugin execution for the same container namespace


aaithal:
This can additionally be used as the `value` for the `ip-address` key in the data store as well

richardpen:
Fixed in #23

PR #23, opened 2017-05-03
Change the value of ipam key-value pair to id that can be more useful instead of timestamps

### LINKED ISSUE #22: Make all the plugins have the same Supported version
`libcni` requires all the plugin has a same cniversion when invoking a list of plugins.

richardpen:
Fix #22 and #5 issue, where ipam plugin will store the id passed in as the value instead of the  timestamps, also added the function that ipam can release the ip address by this id.

Manually tested with both setup and cleanup as in this PR: aws/amazon-ecs-agent/#779

### DIFF
--- a/plugins/ipam/config/config.go
+++ b/plugins/ipam/config/config.go
@@ -44,6 +44,7 @@ type IPAMConfig struct {
 	IPV4Address types.IPNet    `json:"ipv4-address,omitempty"`
 	IPV4Gateway net.IP         `json:"ipv4-gateway,omitempty"`
 	IPV4Routes  []*types.Route `json:"ipv4-routes,omitempty"`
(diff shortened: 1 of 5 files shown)


plugins/ipam/config/config.go
```diff
@@ -107,6 +108,10 @@ func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, string, error) {
 		}
 	}
 
+	if ipamConf.IPAM.ID == "" {
```
aaithal: In the larger scheme of things, should we make this a required field? If an ID is not provided, the worst thing that can happen is that multiple ip addresses will map to same value (blank string). Sounds like we are imposing an additional and unnecessary complexity by making this a required field.

PR #28, opened 2017-05-17
Implement the ecs-bridge plugin

aaithal:
### Details
* plugins/ecs-bridge: add ecs-bridge plugin
* vendor dependencies for the ecs-bridge plugin
* add ecs-bridge target to make file
* pkg, plugins: regenrate mocks for ecs-bridge interfaces

### Testing Done
* [x] Unit tests
* [X]  Manually ran end-to-end using the following config:
```
{
        "cniVersion": "0.3.0",
        "name": "mynet",
        "type": "bridge",
        "bridge": "mybridge",
        "ipam": {
                "type": "ipam",
                "ipv4-subnet": "192.168.3.0/29",
                "ipv4-routes": [
                        {"dst": "169.254.170.2/32"}
                ]
        }
}
```
Output:
```bash
$ sudo CNI_PATH=$CNI_PATH ./docker-run.sh -it busybox
/ # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
3: eth0@if65: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue
    link/ether 0a:58:c0:a8:03:02 brd ff:ff:ff:ff:ff:ff
    inet 192.168.3.2/29 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::5486:5ff:fe97:7fd3/64 scope link
       valid_lft forever preferred_lft forever
/ # ip route
169.254.170.2 via 192.168.3.1 dev eth0
```


aaithal:
This is not a WIP anymore. @nmeyerhans, @samuelkarp, @vsiddharth, and @richardpen can you please review?

plugins/ecs-bridge/engine/createveth.go
```diff
@@ -0,0 +1,66 @@
...
+	// containerInterfaceResult is set when the closure executes. Don't
+	// expect this to be initialized
+	containerInterfaceResult *current.Interface
+}
+
+func newCreateVethPairContext(
+	interfaceName string,
+	mtu int,
+	ip cniipwrapper.IP) *createVethPairContext {
+	return &createVethPairContext{
```
richardpen: please add a newline here, it's a little hard to read.

plugins/ecs-bridge/commands/commands.go
```diff
@@ -0,0 +1,122 @@
...
+	defer log.Flush()
+	return del(args, engine.New())
+}
+
+func add(args *skel.CmdArgs, engine engine.Engine) error {
+	conf, err := types.NewConf(args)
+	if err != nil {
+		// TODO: We log and return errors throughout this function.
+		// Either should be sufficient.
+		log.Errorf("Error loading config from args: %v", err)
```
aaithal: It's being logged into it's own file. However, I do think that the error itself should probably have that context. I'll modify the error messages from types.go to include that.

plugins/ecs-bridge/commands/commands.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	log.Infof("Creating the bridge: %s", conf.BridgeName)
+	bridge, err := engine.CreateBridge(conf.BridgeName, conf.MTU)
+	if err != nil {
```
aaithal: I'm not quite sure what you mean by this. Are you asking for better error messages or more logging?

plugins/ecs-bridge/types/types.go
```diff
@@ -0,0 +1,54 @@
...
+// NewConf creates a new NetConf object by parsing the arguments supplied
+func NewConf(args *skel.CmdArgs) (*NetConf, error) {
+	var conf NetConf
+	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
+		return nil, errors.Wrap(err, "newconf types: failed to parse config")
+	}
+
+	if conf.BridgeName == "" {
+		return nil, errors.Errorf(
+			"newconf types: missing required parameter in config: '%s'",
```
aaithal: The string "bridge", as per the next line

plugins/ecs-bridge/types/types.go
```diff
@@ -0,0 +1,54 @@
...
+			"bridge")
+	}
+
+	if conf.MTU == 0 {
+		// TODO: Is this really needed?
+		conf.MTU = defaultMTU
+	}
+	log.Debugf("Loaded config: %v", conf)
```
aaithal: Because each plugin has its own log file, I don't think that's needed.

plugins/ecs-bridge/main.go
```diff
@@ -0,0 +1,62 @@
...
+	log "github.com/cihub/seelog"
+	"github.com/containernetworking/cni/pkg/skel"
+)
+
+const (
+	defaultLogFilePath = "/var/log/ecs/ecs-cni-bridge-plugin.log"
+)
+
+func init() {
+	runtime.LockOSThread()
```
aaithal: Sure, I'll copy the one from eni plugin

plugins/ecs-bridge/main.go
```diff
@@ -0,0 +1,62 @@
...
+func main() {
+	defer log.Flush()
+	logger.SetupLogger(logger.GetLogFileLocation(defaultLogFilePath))
+
+	var printVersion bool
+	flag.BoolVar(&printVersion, "version", false, "prints version and exits")
+	flag.Parse()
+
+	if printVersion {
+		printVersionInfo()
```
aaithal: That's a good point. Will modify this here and for other plugins.

plugins/ecs-bridge/engine/interfaceip.go
```diff
@@ -0,0 +1,69 @@
...
+package engine
+
+import (
+	"github.com/aws/amazon-ecs-cni-plugins/pkg/netlinkwrapper"
+	"github.com/containernetworking/cni/pkg/ns"
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// getxInterfaceIPV4AddressContext wraps the parameters and the method to
```
aaithal: Sure, will do.

plugins/ecs-bridge/engine/interfaceip.go
```diff
@@ -0,0 +1,69 @@
...
+// get the ipv4 address of the container's veth's interface. This should
+// only be needed if there's a need to determine the IPv4 address of the
+// interface before executing the DEL command using the ipam plugin
+type getInterfaceIPV4AddressContext struct {
```
aaithal: I don't follow? Can you please elaborate. It's being used by engine.go

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/aws/amazon-ecs-cni-plugins/pkg/netlinkwrapper"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// zeroLengthIPString is what we expect net.IP.String() to return if the
+// ip has length 0. We use this to determing if an IP is empty.
+// Refer https://golang.org/pkg/net/#IP.String
+const zeroLengthIPString = "<nil>"
```
aaithal: so, something like:  `net.ParseIP(ip.String())`?

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,245 @@
...
+	bridgeLinkAttributes := netlink.NewLinkAttrs()
+	bridgeLinkAttributes.MTU = mtu
+	bridgeLinkAttributes.Name = bridgeName
+
+	bridge := &netlink.Bridge{
+		LinkAttrs: bridgeLinkAttributes,
+	}
+
+	err := engine.netLink.LinkAdd(bridge)
+	if err != nil && err != syscall.EEXIST {
```
aaithal: Sure. I wanted to modify this like that anyway.

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,245 @@
...
+	hostVethInterface, err := engine.netLink.LinkByName(hostVethName)
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"bridge create veth pair: unable to look up host veth interface %s", hostVethName)
+	}
+
+	err = engine.netLink.LinkSetMaster(hostVethInterface, bridge)
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"bridge create veth pair: unable to assign the veth interface %s to bridge", hostVethName)
```
aaithal: `attach`, will modify

plugins/ecs-bridge/engine/createveth.go
aaithal: done

plugins/ecs-bridge/engine/interfaceip.go
richardpen: This package is used to get the ip address of the veth inside the container namespace, I thought it was only used  if we want the bridge to pass the ip to ipam when delete the ip from ipam. Does it have any other use case I wasn't aware?

plugins/ecs-bridge/engine/engine.go
richardpen: yes, unless you oppose the long statement.

plugins/ecs-bridge/commands/commands.go
richardpen: yes, better error message. if we simply return an error, we may not know where exactly it was generated, like the one `invalid CIDR`.

plugins/ecs-bridge/engine/interfaceip.go
aaithal: I still have the `engine.GetInterfaceIPV4Address` method there in-case we need it in the future. You're right that we don't use this code today. We can delete it later so that we have it in the history if/when we need it in the future.

plugins/ecs-bridge/engine/engine.go
aaithal: We can create a util method for it. I'll create an issue for it and address in a different PR since it's an enhancement anyway.

plugins/ecs-bridge/commands/commands.go
aaithal: It's already being wrapped in `CreateBridge` with that context. Let me know if you see it missing anywhere.

Makefile
```diff
@@ -3,6 +3,7 @@ SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
 ROOT := $(shell pwd)
 LOCAL_ENI_PLUGIN_BINARY=bin/plugins/eni
 LOCAL_IPAM_PLUGIN_BINARY=bin/plugins/ipam
+LOCAL_BRIDGE_PLUGIN_BINARY=bin/plugins/bridge
```
vsiddharth: Should we prefix all our plugins with `ecs-`?

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,245 @@
...
+		Mask: result.IPs[0].Address.Mask,
+	}
+	resultBridgeCIDR := resultBridgeNetwork.String()
+	if len(addrs) > 0 {
+		for _, addr := range addrs {
+			if addr.IPNet.String() == resultBridgeCIDR {
+				return nil
+			}
+		}
+
```
vsiddharth: nit: remove empty line

plugins/ecs-bridge/engine/engine.go
aaithal: Thinking more on this and after playing a bit with the code, I think we need support from `netlink.LinkByName` to do this correctly. We need it to return a named `LinkNotFound"` error instead of `fmt.Errorf()` stuff that it does today. I'll come back to this once we fix that upstream.

Makefile
aaithal: That's happening in a separate PR. Being tracked her https://github.com/aws/amazon-ecs-cni-plugins/issues/38

pkg/cniipamwrapper/ipam.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/containernetworking/cni/pkg/types/current"
+)
+
+// IPAM wraps methods used from the the cni/pkg/ipam package
+type IPAM interface {
+	// ExecAdd invokes the IPAM plugin with the ADD command
+	ExecAdd(plugin string, netconf []byte) (types.Result, error)
+	// ExecDel invokes the IPAM plugin with the ADD command
+	ExecDel(plugin string, netconf []byte) error
+	// ConfifureIface configures the interface named ifName
```
jahkeup: Small typo: `ConfigureIface`.

pkg/cniipwrapper/ip.go
```diff
@@ -0,0 +1,52 @@
...
+
+// IPAM wraps methods used from the the cni/pkg/ip package
+type IP interface {
+	// SetupVeth creates a veth pair
+	SetupVeth(contVethName string, mtu int, hostNS ns.NetNS) (net.Interface, net.Interface, error)
+	// SetHWAddrByIP sets the hardware address for the interface identified
+	// by the ip address
+	SetHWAddrByIP(ifName string, ip4 net.IP, ip6 net.IP) error
+	// DelLinkByNameAddr deletes the interface
+	DelLinkByNameAddr(ifName string, family int) (*net.IPNet, error)
```
liwenwu-amazon: How about "DelIntfByNameAddr(...)"

plugins/ecs-bridge/commands/commands.go
```diff
@@ -0,0 +1,123 @@
...
+		hostVethInterface,
+		// the container veth interface
+		containerVethInterface,
+	}
+
+	// Set the index for the container veth interface in the `Interfaces`
+	// list populated above.
+	// The `ipam.ConfigureIface` method needs this index to be set as it
+	// needs to know which interface should be used when adding routes
+	result.IPs[0].Interface = 2
```
jahkeup: This `2` appears magical even with the above comment, would you consider using a literal (example below) with index for the above or declare the expectation in the comment above the Interfaces declaration?

```go
  result.Interfaces = []*current.Interface{
    // the bridge interface
    0: &current.Interface{
      Name: bridge.Attrs().Name,
      Mac:  bridge.Attrs().HardwareAddr.String(),
    },
    // the host veth interface
    1: hostVethInterface,
    // the container veth interface
    2: containerVethInterface,
  }
```

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+}
+
+// CreateBridge creates the bridge if needed
+func (engine *engine) CreateBridge(bridgeName string, mtu int) (*netlink.Bridge, error) {
+	bridge := &netlink.Bridge{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:   bridgeName,
+			MTU:    mtu,
+			TxQLen: -1,
```
liwenwu-amazon: why "-1"?  can use define a constant for this "-1"?

pkg/cniipamwrapper/ipam.go
samuelkarp: Can we call it an `Interface` instead of `Iface`?

plugins/ecs-bridge/engine/configureveth.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
liwenwu-amazon: can we rename this file as "configure_veth.go"?

pkg/cniipamwrapper/ipam.go
aaithal: @samuelkarp Nope, that's what's its called in `github.com/containernetworking/cni/pkg/ipam`

plugins/ecs-bridge/commands/commands.go
aaithal: sure

pkg/cniipamwrapper/ipam.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/containernetworking/cni/pkg/ipam"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
+)
+
+// IPAM wraps methods used from the the cni/pkg/ipam package
+type IPAM interface {
+	// ExecAdd invokes the IPAM plugin with the ADD command
+	ExecAdd(plugin string, netconf []byte) (types.Result, error)
```
aaithal: Because that's what its called in `github.com/containernetworking/cni/pkg/ipam`

pkg/cniipwrapper/ip.go
aaithal: Nope, because that's what its called in `github.com/containernetworking/cni/pkg/ipam`

plugins/ecs-bridge/engine/configureveth.go
aaithal: I don't see a big value with doing that. It's not that big a string and is composed of 2 distinct parts.

plugins/ecs-bridge/engine/engine.go
aaithal: I've modified this a bit in a later commit where a constructor does that for me. It's basically a setting to do this: https://github.com/vishvananda/netlink/blob/99091d844046e6b17e4b094e937e8e9254afbfbd/link.go#L76

plugins/ecs-bridge/commands/commands_test.go
```diff
@@ -0,0 +1,275 @@
...
+// permissions and limitations under the License.
+
+package commands
+
+import (
+	"errors"
+	"net"
+	"testing"
+
+	"github.com/aws/amazon-ecs-cni-plugins/plugins/ecs-bridge/engine/mocks"
```
jhaynes: Why isn't this in a separate local import block?

plugins/ecs-bridge/engine/engine.go
jhaynes: Worth a TODO or an issue in any case.

plugins/ecs-bridge/commands/commands.go
```diff
@@ -0,0 +1,123 @@
...
+		1: hostVethInterface,
+		// the container veth interface
+		2: containerVethInterface,
+	}
+
+	// Set the index for the container veth interface in the `Interfaces`
+	// list populated above.
+	// The `ipam.ConfigureIface` method needs this index to be set as it
+	// needs to know which interface should be used when adding routes
+	result.IPs[0].Interface = 2
```
aaithal: @jahkeup said:

This `2` appears magical even with the above comment, would you consider using a literal (example below) with index for the above or declare the expectation in the comment above the Interfaces declaration?

```  
result.Interfaces = []*current.Interface{
    // the bridge interface
    0: &current.Interface{
      Name: bridge.Attrs().Name,
      Mac:  bridge.Attrs().HardwareAddr.String(),
    },
    // the host veth interface
    1: hostVethInterface,
    // the container veth interface
    2: containerVethInterface,
  }
```

plugins/ecs-bridge/commands/commands.go
aaithal: You usually wrap the error at the place of its origin. There's no additional context to add here. `CreateBridge` already has enough context within it.

pkg/cniipamwrapper/ipam.go
```diff
@@ -0,0 +1,50 @@
...
+	// ExecDel invokes the IPAM plugin with the ADD command
+	ExecDel(plugin string, netconf []byte) error
+	// ConfifureIface configures the interface named ifName
+	// with the result object
+	ConfigureIface(ifName string, res *current.Result) error
+}
+
+type _ipam struct{}
```
aaithal: `_` is used here because its a dummy wrapper struct and object that wont be used anywhere else

plugins/ecs-bridge/commands/commands_test.go
aaithal: can you elaborate? I don't quite follow the question here.

plugins/ecs-bridge/engine/engine.go
aaithal: This has been modified in the latest commit and we don't do `syscall` stuff anymore.

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,234 @@
...
+	return &engine{
+		netLink: netlinkwrapper.NewNetLink(),
+		ns:      cninswrapper.NewNS(),
+		ip:      cniipwrapper.New(),
+		ipam:    cniipamwrapper.New(),
+	}
+}
+
+// CreateBridge creates the bridge if needed
+func (engine *engine) CreateBridge(bridgeName string, mtu int) (*netlink.Bridge, error) {
```
liwenwu-amazon: Maybe due to poor naming of "netlink/link.go" or lack of comments from this routine, I find this routine is very confusing...

do we have to use vishuvananda/netlink/link.go ?
...

aaithal: Yes, that's the best library I've found for doing this. If you have any specific comments, I can address them either documentation or a code refactor.

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,234 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
liwenwu-amazon: just curious why choose the word "engine"?

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/aws/amazon-ecs-cni-plugins/pkg/netlinkwrapper"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// Engine represents the execution engine for the ECS Bridge plugin.
+// It defines all the operations performed during the execution of the
+// plugin
+type Engine interface {
```
liwenwu-amazon: for a typical bridge API, i think we should add 
- BridgeAddPort(...)
- BridgeDelPort(...)
it is not cleared where these critical bridge actions is called...

plugins/ecs-bridge/engine/engine.go
```diff
@@ -0,0 +1,234 @@
...
+	hostVethName := setupContext.hostVethName
+	containerInterfaceResult := setupContext.containerInterfaceResult
+
+	hostVeth, err := engine.netLink.LinkByName(hostVethName)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err,
+			"bridge create veth pair: unable to look up host veth interface %s", hostVethName)
+	}
+
+	err = engine.netLink.LinkSetMaster(hostVeth, bridge)
```
liwenwu-amazon: I think we should have a API to wrap the poor implementation of netLink package.  for example , we can create a API 
BridgeAddPort(hostVeth, bridge)
and invoke this explicit in command.add(...).

plugins/ecs-bridge/commands/commands.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	log.Infof("Creating veth pair for namespace: %s", args.Netns)
+	containerVethInterface, hostVethInterface, err := engine.CreateVethPair(
+		args.Netns, bridge, conf.MTU, args.IfName)
+	if err != nil {
+		return err
+	}
+
```
liwenwu-amazon: I think it is clear to explicit callout attaching vethHost to bridge here with a new API e.g.
BridgeAddPort(...)

plugins/ecs-bridge/engine/engine.go
aaithal: Because, it's the engine that drives all of the `cni` commands (like `ADD`, `DEL`)

plugins/ecs-bridge/engine/engine.go
aaithal: The workflow that these methods represent are fairly obvious and simple to understand IMO:
1. Create bridge
2. Create veth pair
3. Assign the other end to the container
4. Assign one end of veth pair to the bridge

Between steps 3 & 4, configure the container's veth pair with ip addresses. I would not like to abstract `veth pairs` with the term `port`, because a port on linux can also refer to the port number used for socket communication.

plugins/ecs-bridge/engine/engine.go
aaithal: I think `LinkSetMaster` is fairly obvious without needing our own abstractions. Can you please elaborate what's poor about this implementation?

plugins/ecs-bridge/commands/commands.go
aaithal: Since the package method provided by `cni` already does this for us, I would not be in favor of reimplementing/duplicating that code here.

plugins/ecs-bridge/commands/commands_test.go
jhaynes: Sounds like my memory of goimports rules is fuzzy. Carry on. Nothing to see here.

plugins/ecs-bridge/engine/engine.go
jahkeup: IMHO this particular type name (`Engine`) is severely overloaded and broadly used in a few of our packages and does lack some clarity and precision on the type's purpose.

pkg/cniipamwrapper/ipam.go
jahkeup: If its used only to distinguish this type from the `ipam` that's imported, I'd prefer that we name the import differently (`cni_ipam` was suggested) and use `ipam` for our unexported type name.

plugins/ecs-bridge/engine/engine.go
liwenwu-amazon: IMHO LinkSetMaster() is very bad bridge API.
here are some linux API example on bridge
https://code.woboq.org/linux/linux/net/bridge/br_if.c.html  // linux kernel bridge implementation
or brctlutil,  they all have APIs on
add-bridge
add-interface(or ort)

pkg/cniipwrapper/ip.go
```diff
@@ -0,0 +1,52 @@
...
+	"net"
+
+	"github.com/containernetworking/cni/pkg/ip"
+	"github.com/containernetworking/cni/pkg/ns"
+)
+
+// IPAM wraps methods used from the the cni/pkg/ip package
+type IP interface {
+	// SetupVeth creates a veth pair
+	SetupVeth(contVethName string, mtu int, hostNS ns.NetNS) (net.Interface, net.Interface, error)
```
jahkeup: [The original docstring](https://github.com/containernetworking/cni/blob/c71a7c068fd7f88a84e5dc0f9b0258f612cc1e68/pkg/ip/link.go#L117-L120) gives more clarity. Can we reference this in some way or copy it in place for the interface'd functions?

plugins/ecs-bridge/engine/engine.go
liwenwu-amazon: cool,  I thought  "step 4" is inside "step2"  CreateVethPair()   by LinkSetMaster(),   is my understanding correct?

pkg/cniipamwrapper/ipam.go
aaithal: alright, sure

plugins/ecs-bridge/engine/engine.go
aaithal: If you have better alternatives, i'm all ears

plugins/ecs-bridge/engine/engine.go
aaithal: ok, i'd still not like to abstract or wrap something just because its name is not the same as in some other implementation.

plugins/ecs-bridge/engine/engine.go
aaithal: No, `CreateVethPair` doesn't invoke `LinkSetMaster`. That's happening in `AttachHostVethInterfaceToBridge`

liwenwu-amazon: cool!,  When there are multiple Name Spaces, will there be ONE bridge share by all ? or One bridge per name space?

aaithal: > cool!, When there are multiple Name Spaces, will there be ONE bridge share by all ?

Yes, the bridge itself resides in the default namespace and is used to connect all container namespaces to the default namespace so that they can query the Agent for credentials.

aaithal:
@richardpen I accidentally merged this PR. Apologies for that. Please add any additional comments that you might have and i'll address them in a new PR.

plugins/ecs-bridge/engine/engine.go
jahkeup: How about `BridgePlugin` or even `Plugin`? The actions taken are of the plugin itself so it makes some sense and is slightly more contextual than Engine (I'm not the greatest at coming up with names).

aaithal: We could actually do better and break this up into:
```
type Bridge interface {
Create()
CreateVeth()
AttachHostVethInterface()
Configure()
}

type RunIPAM interface {
Add()
Del()
}

type Veth interface {
ConfigureContainer()
GetInterfaceIPV4Address()
Delete()
}
```

But, I don't want to do it right now and can create an issue to follow it up if you like this refactor.

Makefile
```diff
@@ -3,6 +3,7 @@ SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
 ROOT := $(shell pwd)
 LOCAL_ENI_PLUGIN_BINARY=bin/plugins/eni
 LOCAL_IPAM_PLUGIN_BINARY=bin/plugins/ipam
+LOCAL_BRIDGE_PLUGIN_BINARY=bin/plugins/ecs-bridge
 GIT_PORCELAIN=$(shell git status --porcelain | wc -l)
 GIT_SHORT_HASH=$(shell git rev-parse --short HEAD)
 VERSION=$(shell cat $(ROOT)/VERSION)
```
richardpen: All the plugins are sharing the same version, I think each plugin should have its own version, right?

Makefile
```diff
@@ -15,7 +16,7 @@ get-deps:
 	go get github.com/tools/godep
 
 .PHONY: plugins
-plugins: $(LOCAL_ENI_PLUGIN_BINARY) $(LOCAL_IPAM_PLUGIN_BINARY)
+plugins: $(LOCAL_ENI_PLUGIN_BINARY) $(LOCAL_IPAM_PLUGIN_BINARY) $(LOCAL_BRIDGE_PLUGIN_BINARY)
```
richardpen: line 9, All the plugins are sharing the same version, I think each plugin should have its own version, right?

Makefile
```diff
@@ -3,6 +3,7 @@ SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
 ROOT := $(shell pwd)
 LOCAL_ENI_PLUGIN_BINARY=bin/plugins/eni
 LOCAL_IPAM_PLUGIN_BINARY=bin/plugins/ipam
+LOCAL_BRIDGE_PLUGIN_BINARY=bin/plugins/bridge
 GIT_PORCELAIN=$(shell git status --porcelain | wc -l)
 GIT_SHORT_HASH=$(shell git rev-parse --short HEAD)
 VERSION=$(shell cat $(ROOT)/VERSION)
```
aaithal: I think the argument was that we'll increment this version for changes to any of the plugins since we release all of them together (because they are part of the same repo). I think it's much simpler from a maintenance aspect to keep all of them at the same version rather than managing dependencies between them. If you have better suggestions to do this, please add them here and we can adapt to that model

Makefile
aaithal: See response to your comment above.

Issue #26, opened 2017-05-17
IPAM Plugin: IP Address should not be a required field in config for DEL command


aaithal:
We should make the `ipv4-address` field optional for `DEL`. The plugin should accept either that or the unique ID and infer `ipv4-address` from unique ID. Additionally, the plugin should also ensure that when unique id is specified, it is actually unique and that multiple IP addresses do not point to the same ID in the key value store

richardpen:
Merged #29

PR #29, opened 2017-05-18
Added function for ipam to release ip by unique id

richardpen:
Add a function to release the ip from ipam by unique id associated with the ip. Addressing #26 
make unit-test
make integration-test

### DIFF
--- a/plugins/ipam/commands/commands.go
+++ b/plugins/ipam/commands/commands.go
@@ -84,6 +84,10 @@ func Del(args *skel.CmdArgs) error {
 		return err
 	}
 
+	if err := validateDelConfiguration(ipamConf); err != nil {
+		return err
+	}
+
 	dbConf, err := config.LoadDBConfig()
 	if err != nil {
 		return err
@@ -101,18 +105,35 @@ func Del(args *skel.CmdArgs) error {
 	return del(ipManager, ipamConf)
 }
 
-func del(ipManager ipstore.IPAllocator, ipamConf *config.IPAMConfig) error {
-	if ipamConf.IPV4Address.IP == nil {
-		return errors.New("del commands: ip address is required for deletion")
+// validateDelConfiguration checks the configuration for ipam del
+func validateDelConfiguration(ipamConf *config.IPAMConfig) error {
+	if ipamConf.ID == "" && (ipamConf.IPV4Address.IP == nil || net.ParseIP(ipamConf.IPV4Address.IP.String()) == nil) {
+		return errors.New("del commands: ip address and id can not both be empty for deletion")
 	}
+	return nil
+}
 
-	err := ipManager.Release(ipamConf.IPV4Address.IP.String())
-	if err != nil {
-		return err
+func del(ipManager ipstore.IPAllocator, ipamConf *config.IPAMConfig) error {
+	var releasedIP string
+
+	if ipamConf.IPV4Address.IP != nil && net.ParseIP(ipamConf.IPV4Address.IP.String()) != nil {
+		// Release the ip by ip address
+		err := ipManager.Release(ipamConf.IPV4Address.IP.String())
+		if err != nil {
+			return err
+		}
+		releasedIP = ipamConf.IPV4Address.IP.String()
+	} else {
+		// Release the ip by unique id associated with the ip
+		ip, err := ipManager.ReleaseByID(ipamConf.ID)
+		if err != nil {
+			return nil
+		}
+		releasedIP = ip
 	}
 
 	// Update the last known ip
-	err = ipManager.Update("lastKnownIP", ipamConf.IPV4Address.IP.String())
+	err := ipManager.Update("lastKnownIP", releasedIP)
 	if err != nil {
 		// This error will only impact how the next ip will be find, it shouldn't cause
 		// the command to fail
--- a/plugins/ipam/ipstore/ipstore.go
+++ b/plugins/ipam/ipstore/ipstore.go
@@ -18,7 +18,6 @@ import (
 	"net"
 	"time"
 
-	log "github.com/cihub/seelog"
 	"github.com/docker/libkv/store"
 	"github.com/docker/libkv/store/boltdb"
 	"github.com/pkg/errors"
@@ -27,6 +26,9 @@ import (
 const (
 	// Mask size beyond 30 won't be accepted since the .0 and .255 are reserved
 	MaxMask = 30
+	// This assumes all the ip start with 1 for now before we address
+	// the issue: https://github.com/aws/amazon-ecs-cni-plugins/issues/37
+	IPPrefix = "1"
 )
 
 // IPManager is responsible for managing the ip addresses using boltdb
@@ -52,6 +54,7 @@ type IPAllocator interface {
 	Assign(string, string) error
 	Update(string, string) error
 	Release(string) error
+	ReleaseByID(string) (string, error)
 	Exists(string) (bool, error)
 	SetLastKnownIP(net.IP)
 	Close()
@@ -92,8 +95,7 @@ func (manager *IPManager) GetAvailableIP(id string) (string, error) {
 		}
 		err = manager.Assign(nextIP.String(), id)
 		if err != nil && err != store.ErrKeyExists {
-			log.Debugf("query to the db failed, err: %v", err)
-			return "", err
+			return "", errors.Wrapf(err, "ipstore get ip: failed to assign ip in ipam")
 		} else if err == store.ErrKeyExists {
 			// ip already be used
 		} else {
@@ -129,6 +131,18 @@ func (manager *IPManager) Assign(ip string, id string) error {
 	if ok {
 		return store.ErrKeyExists
 	}
+
+	// if the id presents, it should be unique
+	if id != "" {
+		ok, err := manager.UniqueID(id)
+		if err != nil {
+			return errors.Wrapf(err, "assign ipstore: check id unique failed, id: %s", id)
+		}
+		if !ok {
+			return errors.Errorf("assign ipstore: id already exists, id: %s", id)
+		}
+	}
+
 	err = manager.client.Put(ip, []byte(id), nil)
 	if err != nil {
 		return errors.Wrapf(err, "assign ipstore: failed to put the key/value into the db: %s -> %s", ip, id)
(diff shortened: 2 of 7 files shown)


plugins/ipam/commands/commands.go
```diff
@@ -100,17 +100,28 @@ func Del(args *skel.CmdArgs) error {
...
 func del(ipManager ipstore.IPAllocator, ipamConf *config.IPAMConfig) error {
-	if ipamConf.IPV4Address.IP == nil {
-		return errors.New("del commands: ip address is required for deletion")
-	}
+	var releasedIP string
 
-	err := ipManager.Release(ipamConf.IPV4Address.IP.String())
-	if err != nil {
-		return err
+	if ipamConf.IPV4Address.IP != nil {
```
aaithal: I think `ipamConf.IPV4Address.IP != nil && ipamConf.IPV4Address.IP.String() == "<nil>"` is a stronger check since this is [not a pointer type](https://github.com/richardpen/amazon-ecs-cni-plugins-1/blob/eae4a48d251e6481aedaa7ec9f6564efdbad5f33/plugins/ipam/config/config.go#L40).

plugins/ipam/commands/commands.go
```diff
@@ -100,17 +100,28 @@ func Del(args *skel.CmdArgs) error {
...
+		// Release the ip by ip address
+		err := ipManager.Release(ipamConf.IPV4Address.IP.String())
+		if err != nil {
+			return err
+		}
+		releasedIP = ipamConf.IPV4Address.IP.String()
+	} else if ipamConf.ID != "" {
```
aaithal: I think `ipamConf.ID != nil && ipamConf.ID == ""` is a stronger check since this is [not a pointer type](https://github.com/richardpen/amazon-ecs-cni-plugins-1/blob/eae4a48d251e6481aedaa7ec9f6564efdbad5f33/plugins/ipam/config/config.go#L47).

plugins/ipam/ipstore/ipstore.go
```diff
@@ -52,6 +51,7 @@ type IPAllocator interface {
 	Assign(string, string) error
 	Update(string, string) error
 	Release(string) error
+	ReleaseByValue(string) (string, error)
```
aaithal: Can you rename this as `ReleaseByID` since that's the abstraction provided by this interface?

plugins/ipam/ipstore/ipstore.go
```diff
@@ -92,8 +92,7 @@ func (manager *IPManager) GetAvailableIP(id string) (string, error) {
 		}
 		err = manager.Assign(nextIP.String(), id)
 		if err != nil && err != store.ErrKeyExists {
-			log.Debugf("query to the db failed, err: %v", err)
-			return "", err
+			return "", errors.Wrapf(err, "failed to assign ip in ipam")
```
aaithal: Can you reformat this as `"ipstore get ip:  failed to assign ip in ipam"`

plugins/ipam/ipstore/ipstore.go
```diff
@@ -129,6 +128,15 @@ func (manager *IPManager) Assign(ip string, id string) error {
 	if ok {
 		return store.ErrKeyExists
 	}
+
+	// if the id is presented, it should be unique
+	if id != "" {
+		err := manager.UniqueID(id)
+		if err != nil {
+			return errors.Wrapf(err, "assign ipstore: check id unique failed")
```
aaithal: Please include the `id` in error message as well.

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
 	return nil
 }
 
+// ReleaseByValue release the key-value pair by value
+func (manager *IPManager) ReleaseByValue(value string) (string, error) {
+	// TODO improve this part by implement listing all the kv pairs
```
aaithal: Can you create an issue for this as well?

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
...
+	// libkv library only provide list all key-value pairs with prefix of key, and
+	// if the result is empty it will return store.ErrKeyNotFound error
+	kvPairs, err := manager.client.List("1")
```
aaithal: What is the meaning of this string (`"1"`)? Can you const this and add a comment explaining the same?

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
...
+	if err == store.ErrKeyNotFound {
+		return "", errors.Errorf("releaseByValue ipstore: no ip associated with the id: %s", value)
```
aaithal: `"release ipstore"` is more user friendly i think.

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
...
+	}
+	if err != nil {
+		return "", errors.Wrapf(err, "releaseByValue ipstore: failed to list the key-value pairs in db")
```
aaithal: `"release ipstore"` is more user friendly i think.

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
...
+		if string(kvPair.Value) == value {
+			err = manager.Release(kvPair.Key)
+			if err != nil {
+				return "", err
+			}
+			ipDeleted = kvPair.Key
+			return ipDeleted, nil
+		}
+	}
+	return "", errors.Errorf("No ip associated with the given value: %s", value)
```
aaithal: Can you reformat the error message as `"release ipstore: no ip address associated ..."`

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +167,53 @@ func (manager *IPManager) Release(ip string) error {
...
+}
+
+// UniqueID checks whether the id has already existed in the ipam
+func (manager *IPManager) UniqueID(id string) error {
```
aaithal: This should return `(bool, error)`. I think we are conflating the two by returning just `error`

plugins/ipam/commands/commands.go
richardpen: ID is a string type, it shouldn't be nil.

plugins/ipam/ipstore/ipstore.go
```diff
@@ -129,6 +131,18 @@ func (manager *IPManager) Assign(ip string, id string) error {
 	if ok {
 		return store.ErrKeyExists
 	}
+
+	// if the id is presented, it should be unique
```
jhaynes: nit, s/presented/present/

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +173,53 @@ func (manager *IPManager) Release(ip string) error {
...
 }
 
+// ReleaseByID release the key-value pair by id
+func (manager *IPManager) ReleaseByID(id string) (string, error) {
+	// TODO improve this part by implement listing all the kv pairs
+	// libkv library only provide list all key-value pairs with prefix of key, and
+	// if the result is empty it will return store.ErrKeyNotFound error
+	kvPairs, err := manager.client.List(IPPrefix)
+	if err == store.ErrKeyNotFound {
+		return "", errors.Errorf("release ipstore: no ip associated with the id: %s", id)
```
jhaynes: Why `errors.Errorf` here instead of `errors.Wrapf`? We're sure we don't care about `store.ErrKeyNotFound`'s context?

richardpen: In this case `store.ErrKeyNotFound` isn't really an error, it's just because today we don't have a better way to get all the key-value pairs from the db. If we can fix #37 then we will get rid of this error.

plugins/ipam/commands/commands.go
```diff
@@ -100,17 +100,29 @@ func Del(args *skel.CmdArgs) error {
...
 func del(ipManager ipstore.IPAllocator, ipamConf *config.IPAMConfig) error {
-	if ipamConf.IPV4Address.IP == nil {
-		return errors.New("del commands: ip address is required for deletion")
-	}
+	var releasedIP string
 
-	err := ipManager.Release(ipamConf.IPV4Address.IP.String())
-	if err != nil {
-		return err
+	if ipamConf.ID == "" && (ipamConf.IPV4Address.IP == nil || net.ParseIP(ipamConf.IPV4Address.IP.String()) == nil) {
```
aaithal: Can you move this block to `Del()` and refactor  config parsing part of `Del()` into a separate method?

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +173,53 @@ func (manager *IPManager) Release(ip string) error {
...
+		return true, nil
+	}
+
+	if err != nil {
+		return false, errors.Wrapf(err, "release ipstore: failed to list the key-value pairs in db")
+	}
+
+	for _, kvPair := range kvPairs {
+		if string(kvPair.Value) == id {
+			return false, errors.Errorf("UniqueID ipstore: id already exists: %s", id)
```
aaithal: I don't think we need `UniqueID` string in the error message.

plugins/ipam/ipstore/ipstore.go
```diff
@@ -159,6 +173,53 @@ func (manager *IPManager) Release(ip string) error {
...
+// UniqueID checks whether the id has already existed in the ipam
+func (manager *IPManager) UniqueID(id string) (bool, error) {
+	kvPairs, err := manager.client.List(IPPrefix)
+	// TODO improve this part by implement listing all the kv pairs
+	if err == store.ErrKeyNotFound {
+		return true, nil
+	}
+
+	if err != nil {
+		return false, errors.Wrapf(err, "release ipstore: failed to list the key-value pairs in db")
```
aaithal: I don't think you need `release ` here. Just `ipstore: failed..` should be sufficient

Issue #21, opened 2017-04-13
Make ENI plugin return consistent result as other plugin 


richardpen:
Make `Add` and `Del` command in eni plugin return a [`Result`](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L136), to get rid of the error message caused by [`json.Unmarshal`](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L31).

aaithal:
Fixed with #30

PR #30, opened 2017-05-19
make eni return the consistent results as cni

richardpen:
Address issue #21 
make unit-test
make integration-test
Manually tested, no error will printed.

### DIFF
--- a/plugins/eni/commands/commands.go
+++ b/plugins/eni/commands/commands.go
@@ -15,11 +15,15 @@ package commands
 
 import (
 	"fmt"
+	"net"
 
 	"github.com/aws/amazon-ecs-cni-plugins/plugins/eni/engine"
 	"github.com/aws/amazon-ecs-cni-plugins/plugins/eni/types"
+
 	log "github.com/cihub/seelog"
 	"github.com/containernetworking/cni/pkg/skel"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/pkg/errors"
 )
 
@@ -78,6 +82,18 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 	}
 	log.Debugf("Found ipv4 gateway and netmask for ENI: %s %s", ipv4Gateway, ipv4Netmask)
 
+	ipv4Address := fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask)
+	_, ipv4Net, err := net.ParseCIDR(ipv4Address)
+	if err != nil {
+		return errors.Wrapf(err, "add eni: failed to parse ipv4 gateway netmask: %s", fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask))
+	}
+	ips := []*current.IPConfig{
+		{
+			Version: "4",
+			Address: *ipv4Net,
+		},
+	}
+
 	ipv6Address := ""
 	ipv6Gateway := ""
 	if conf.IPV6Address != "" {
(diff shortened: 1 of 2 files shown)


plugins/eni/commands/commands.go
```diff
@@ -110,7 +113,27 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 	}
 	log.Infof("ENI %s has been assigned to the container's namespace", conf.MACAddress)
 
-	return nil
+	_, ipv4Net, err := net.ParseCIDR(ipv4Gateway + "/" + ipv4Netmask)
```
vsiddharth: I think we are better off using `Sprintf` here

plugins/eni/commands/commands.go
```diff
@@ -110,7 +113,27 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
...
+				Name: networkDeviceName,
+				Mac:  macAddressOfENI,
+			},
+		},
+		IPs: []*current.IPConfig{
+			{
+				Version: "4",
+				Address: *ipv4Net,
+			},
+		},
```
vsiddharth: Should we also stick the ipv6 address here ?

plugins/eni/commands/commands.go
```diff
@@ -78,6 +81,17 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 	}
 	log.Debugf("Found ipv4 gateway and netmask for ENI: %s %s", ipv4Gateway, ipv4Netmask)
 
+	_, ipv4Net, err := net.ParseCIDR(fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask))
+	if err != nil {
+		return errors.Wrapf(err, "add eni: parse cidr failed: %s", fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask))
```
aaithal: Can you please add more context to this error message around where it's generated (or make it more customer friendly)?

`"add eni: failed to parse gateway netmask: %s"`

plugins/eni/commands/commands.go
```diff
@@ -97,6 +111,15 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 			return err
 		}
 		log.Debugf("IPV6 Gateway IP: %v", ipv6Gateway)
+		_, ipv6net, err := net.ParseCIDR(ipv6Address)
+		if err != nil {
+			return errors.Wrapf(err, "parsing ipv6 failed:%s", ipv6Address)
```
aaithal: Same here, please make provide more context here: 

`"add eni: failed to parse ipv6 gateway: %s"`

plugins/eni/commands/commands.go
```diff
@@ -15,11 +15,14 @@ package commands
...
 import (
 	"fmt"
+	"net"
 
 	"github.com/aws/amazon-ecs-cni-plugins/plugins/eni/engine"
 	"github.com/aws/amazon-ecs-cni-plugins/plugins/eni/types"
 	log "github.com/cihub/seelog"
 	"github.com/containernetworking/cni/pkg/skel"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
```
vsiddharth: Nit: We should also separate out the 3p imports by adding a newline

plugins/eni/commands/commands.go
```diff
@@ -78,6 +81,17 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 	}
 	log.Debugf("Found ipv4 gateway and netmask for ENI: %s %s", ipv4Gateway, ipv4Netmask)
 
+	_, ipv4Net, err := net.ParseCIDR(fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask))
```
vsiddharth: Nit: store `fmt.Sprintf("%s/%s", ipv4Gateway, ipv4Netmask)` as a string and make both ipv4 and ipv6 checks semantically consistent

plugins/eni/commands/commands_test.go
```diff
@@ -15,12 +15,18 @@ package commands
 
 import (
 	"errors"
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/aws/amazon-ecs-cni-plugins/plugins/eni/engine/mocks"
```
jhaynes: Similar to @vsiddharth's Nit above: this should be in its own local import "chunk" shouldn't it?

PR #32, opened 2017-05-22
Get rid of loading configuration from arguments

### LINKED ISSUE #31: `ipam` plugin could potentially panic when set the parameter from args
In `ipam` plugin it calls [`types.LoadArgs`](https://github.com/aws/amazon-ecs-cni-plugins/blob/dev/plugins/ipam/config/config.go#L70) to load configuration from the `args` which could cause panic if the configuration field didn't implement the `TextUnmarshaler` interface, see [here](https://github.com/containernetworking/cni/blob/master/pkg/types/args.go#L89).

richardpen:
Fix #31 

* make plugins
* make unit-test
* make integration-test
* Manual test with the built binary


vsiddharth:
LGTM 👍

PR #40, opened 2017-05-26
flush logs in add and delete commands.

aaithal:
Doing this in main doesn't capture the  logs that are emitted in add and delete commands.

### Testing Done
- [X] Build succeeds
- [X] Unit tests pass

### DIFF
--- a/plugins/ipam/commands/commands.go
+++ b/plugins/ipam/commands/commands.go
@@ -28,6 +28,7 @@ import (
 // Add will return ip, gateway, routes which can be
 // used in bridge plugin to configure veth pair and bridge
 func Add(args *skel.CmdArgs) error {
+	defer seelog.Flush()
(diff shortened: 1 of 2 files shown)


plugins/ipam/commands/commands.go
```diff
@@ -28,6 +28,7 @@ import (
 // Add will return ip, gateway, routes which can be
 // used in bridge plugin to configure veth pair and bridge
 func Add(args *skel.CmdArgs) error {
+	defer seelog.Flush()
```
vsiddharth: Can we switch the alias from `seelog` to `log` and make it consistent across all our packages ?

aaithal: Even if we do that here, it's hard to enforce it across all packages. I'll leave it be.

PR #44, opened 2017-06-02
Implement dhclient path overrides

aaithal:
### Summary
* plugins/eni/engine: enable env var overrides for various dhclient command parameters
* plugins/eni/engine, pkg/oswrapper: regenerate mocks
* plugins/eni: add unit tests for dhclient changes

### Testing Done
* [X] `make clean unit-tests plugins`
* [ ] Manual Testing


plugins/eni/engine/dhclient.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	log.Debugf("dhclient found in: %s", dhclientPath)
+	return true
+}
+
+func (client *dhclient) Start(deviceName string, ipRev int) error {
+	args := client.constructDHClientArgs(deviceName, ipRev)
+	cmd := client.exec.Command(client.executable, args...)
+	out, err := cmd.CombinedOutput()
```
aaithal: @nmeyerhans Thanks for pointing this out. You're right that we do not have enforce any timeouts here beyond what's set as the default. I tried various options mentioned the man page for `dhclient`, but was unable to make the process say in foreground and exit after a timeout interval. I think that enforcing a timeout outside of the `dhclient` command via the invocation code might be an acceptable solution here where we mark it as an error if the timeout is breached. I am in favor of doing this in another PR though. Let me know if that sounds ok to you.

plugins/eni/engine/dhclient.go
```diff
@@ -0,0 +1,283 @@
...
+	if envVal == "" {
+		return defaultValue
+	}
+	return envVal
+}
+
+func (client *dhclient) IsExecutableInPath() bool {
+	dhclientPath, err := client.exec.LookPath(client.executable)
+	if err != nil {
+		log.Warnf("Error searching dhclient in PATH: %v", err)
```
aaithal: Fixed.

PR #45, opened 2017-06-26
Added "--capabilities" command and make version command fail when there is an error

richardpen:
Tested with simple code:
```
./bin/plugins/ecs-ipam --version
{"version":"2017.06.0","dirty":true,"gitShortHash":"226db36"}
 ./bin/plugins/ecs-ipam --capabilities
{"capabilities":["awsvpc-network-mode"]}
```


plugins/ecs-bridge/main.go
```diff
@@ -47,8 +47,7 @@ func main() {
 
 	if printVersion {
 		if err := printVersionInfo(); err != nil {
-			os.Stderr.WriteString(
-				fmt.Sprintf("Error getting version string: %s", err.Error()))
+			fmt.Printf("Error getting version string: %s", err.Error())
```
vsiddharth: `Printf` writes to `stdout` while the original wrote to `stderr`. 
Is this intentional ?

richardpen: yes, write to the stderr won't be visible to the [err](https://github.com/aws/amazon-ecs-agent/blob/enis/agent/ecscni/plugin.go#L185). Write to the stdout will makes it captured by the err.

samuelkarp: You can definitely retrieve `stderr` from processes launched by Go.  I'd prefer to have errors on `stderr`.

aaithal: Would prefer to revert to `os.Stderr` here.

plugins/eni/main.go
```diff
@@ -47,10 +47,9 @@ func main() {
 
 	if printVersion {
 		if err := printVersionInfo(); err != nil {
-			os.Stderr.WriteString(
-				fmt.Sprintf("Error getting version string: %s", err.Error()))
+			fmt.Printf("Error getting version string: %s", err.Error())
```
aaithal: Same as above. Can you revert to using `os.Stderr.WriteString` ?

plugins/ipam/main.go
```diff
@@ -40,10 +40,9 @@ func main() {
 
 	if printVersion {
 		if err := printVersionInfo(); err != nil {
-			os.Stderr.WriteString(
-				fmt.Sprintf("Error getting version string: %s", err.Error()))
+			fmt.Printf("Error getting version string: %s", err.Error())
```
aaithal: Same as above. Can you revert to using `os.Stderr.WriteString` ?

PR #46, opened 2017-07-28
End-to-end tests

aaithal:
## Summary 
Added end-to-end tests for the ECS Bridge, IPAM and ENI Plugins

## Testing
- [X] build (`make release`)
- [X] unit test (`make unit-test`)
- [X] integration test (`make integ-test`)
- [X] end-to-end test (`make e2e-test`)


vsiddharth:
Had a quick look and the changes look good to me. 
Will have a detailed look later. 

However, I think it would be nice to see additional comments in the test to easily understand the flow without having to consult the README.

aaithal:
@vsiddharth done

plugins/eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,334 @@
...
+		err := invoke.ExecPluginWithoutResult(
+			eniPluginPath,
+			netConf,
+			execInvokeArgs)
+		require.NoError(t, err, "Unable to execute ADD command for ecs-eni plugin")
+		return nil
+	})
+
+	targetNS.Do(func(ns.NetNS) error {
+		// Validate that only 2 devices exist in the target network
```
richardpen: do you also want to verify the dhclient is running?

plugins/ecs-bridge/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### End-to-end Tests
+
+The end-to-end test suite for this package makes the following assumptions:
+1. The `ecs-bridge` and `ecs-ipam` plugin executables have been built
+2. The `CNI_PATH` environment variable points to the location of these plugins
+3. The test is being executed with `root` user privileges
+
+Since these tests invoke the the ECS Bridge plugin as if an end user such as 
+the ECS Agent is invoking it, additional configuration variables can be set to
```
aaithal: sure.

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,326 @@
...
+	require.NoError(t, err, "Unable to create directory for storing test logs")
+
+	// Configure the env var to use the test logs directory
+	os.Setenv("ECS_CNI_LOG_FILE", fmt.Sprintf("%s/bridge.log", testLogDir))
+	t.Logf("Using %s for test logs", testLogDir)
+	defer os.Unsetenv("ECS_CNI_LOG_FILE")
+
+	// Handle deletion of test logs at the end of the test execution if
+	// specified
+	ok, err := strconv.ParseBool(getEnvOrDefault("ECS_PRESERVE_E2E_TEST_LOGS", "false"))
```
aaithal: I don't know how to do that with golang's test framework. Let me know if you have any details/info on how to do that. Besides, you can always override that env var while running tests if it fails the first time.

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,326 @@
...
+			addressFound = true
+		}
+	}
+	require.True(t, addressFound, "IP address '%s' not associated with: %s",
+		expectedVethAddress, veth.Attrs().Name)
+}
+
+// validateRouteForVethInTargetNetNS validates that the expected route has been
+// added for the veth device in target network namespace
+func validateRouteForVethInTargetNetNS(t *testing.T, veth netlink.Link) {
```
aaithal: good point. will add that logic.

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,326 @@
...
+		expectedGateway, veth.Attrs().Name)
+}
+
+// validateLinkDoesNotExist validates that the named link does not exist in the
+// network namespace
+func validateLinkDoesNotExist(t *testing.T, name string) {
+	_, err := netlink.LinkByName(name)
+	require.Error(t, err, "Link %s should not exist", name)
+	_, ok := err.(netlink.LinkNotFoundError)
+	require.True(t, ok, "Error type is incorrect for link: %s", name)
```
aaithal: sure

plugins/eni/README.md
```diff
@@ -0,0 +1,80 @@
+# ECS Bridge Plugin
```
aaithal: yeah, you paste what you copy.

plugins/eni/README.md
```diff
@@ -0,0 +1,80 @@
...
+   1. ec2:DescribeSecurityGroups
+   2. ec2:CreateNetworkInterface
+   3. ec2:DescribeNetworkInterfaces
+   4. ec2:DeleteNetworkInterface
+   5. ec2:AttachNetworkInterface
+   6. ec2:DetachNetworkInterface
+3. The EC2 Instance has room to attach at least one ENI. Please refer to 
+[aws eni documentation](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI) for details on the limits based on instance type
+4. The `ecs-eni` plugin executable has been built
+5. The `CNI_PATH` environment variable points to the location of these plugins
```
aaithal: Yeah, if you're running outside of `make`, it's still useful information

plugins/eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/containernetworking/cni/pkg/ns"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/vishvananda/netlink"
+)
+
+const (
+	ifName        = "ecs-test-eth0"
+	containerID   = "contain-er"
```
aaithal: I think CNI requires you to still set it even if unused.

plugins/eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,334 @@
...
+	require.NoError(t, err, "Unable to create ENI")
+	defer deleteENI(ec2Client, eni)
+
+	require.NoError(t, waitUntilNetworkInterfaceAvailable(ec2Client, eni), "ENI didn't transition into 'available'")
+	// Attach the ENI to the instance
+	attachment, err := attachENI(ec2Client, cfg, eni)
+	require.NoError(t, err, "Unable to attach ENI")
+	defer detachENI(ec2Client, attachment)
+
+	waitUntilNetworkInterfaceAttached(eni, 5*time.Second)
```
aaithal: yeah, sure.

plugins/eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,334 @@
...
+	ok, err := strconv.ParseBool(getEnvOrDefault("ECS_PRESERVE_E2E_TEST_LOGS", "false"))
+	assert.NoError(t, err, "Unable to parse ECS_PRESERVE_E2E_TEST_LOGS env var")
+	if !ok {
+		defer os.RemoveAll(testLogDir)
+	}
+
+	// Use the current network namespace to execute the test in
+	testNS, err := ns.GetCurrentNS()
+	require.NoError(t, err, "Unable to get the network namespace to run the test in")
+	defer testNS.Close()
```
aaithal: Yeah, all it does is close the fd's if opened. I don't think it hurts.

plugins/ecs-bridge/e2eTests/e2e_test.go
richardpen: Probably you can use this `Failed()` [method](https://golang.org/pkg/testing/#B.Failed) in the defer function. like this:
```
defer func(){
    if !t.Failed(){
       os.RemoveAll(testLogDir)
    }
}()
```

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,326 @@
...
+	    {
+		"dst":"%s"
+	    }
+	]
+    }
+}`
+)
+
+func init() {
+	runtime.LockOSThread()
```
petderek: Why do we need this? Add a comment?

plugins/ecs-bridge/e2eTests/e2e_test.go
petderek: Is this something we can do at the package level instead of within the test?

plugins/eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,334 @@
...
+    "type":"ecs-eni",
+    "cniVersion":"0.3.0",
+    "eni":"%s",
+    "ipv4-address":"%s",
+    "mac":"%s"
+}`
+)
+
+func init() {
+	runtime.LockOSThread()
```
petderek: Also here. :) Why?

plugins/ecs-bridge/e2eTests/e2e_test.go
aaithal: There's a comment in main.go for why this is needed. I'll copy it here as well.

plugins/ecs-bridge/e2eTests/e2e_test.go
aaithal: @richardpen thanks. let me see what I can do with this.

plugins/eni/e2eTests/e2e_test.go
aaithal: Yes, but there's no sane way of doing that. We can read the pid value from the file as stop-gap, but even that's not super clean I feel. I'll add a TODO as placeholder until we figure out a way to do that.

aaithal:
@petderek @richardpen I've modified the code to address PR comments. Can you take another look?

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, route := range routes {
+		if route.Gw == nil {
+			defaultRouteFound = true
+		} else if route.Dst.String() == dst &&
+			route.Src == nil &&
+			route.Gw.String() == expectedGateway {
+			routeFound = true
+		}
+	}
+	require.True(t, defaultRouteFound,
```
richardpen: shouldn't this be `require.False` as we don't expect the default route. Also this is same to ipv6, so I think you can use `netlink.FAMILY_ALL` instead of just `netlink.FAMILY_V4`. To be simple, there should be only one route existed which is the one we added in the configure file.

aaithal: oops, done.

plugins/ecs-bridge/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,326 @@
+// +build e2e
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aaithal: yes, done.

Makefile
```diff
@@ -7,6 +7,7 @@ LOCAL_BRIDGE_PLUGIN_BINARY=bin/plugins/ecs-bridge
 GIT_PORCELAIN=$(shell git status --porcelain | wc -l)
 GIT_SHORT_HASH=$(shell git rev-parse --short HEAD)
 VERSION=$(shell cat $(ROOT)/VERSION)
+GO_EXECUTABLE=$(shell which go)
```
nmeyerhans: Minor, but `which` is actually not guaranteed to be present. It's not, for example, on the Amazon Linux base docker image, and it's not required by POSIX. We should use `command -v`, which is required by POSIX and thus more compatible. Alternatively, document a dependency on `/bin/sh` being `bash` so we can make use of the built-in.

Makefile
```diff
@@ -46,13 +47,16 @@ $(LOCAL_BRIDGE_PLUGIN_BINARY): $(SOURCES)
...
-.PHONY: unit-test integration-test
+.PHONY: unit-test integration-test e2e-test
 unit-test: $(SOURCES)
 	go test -v -cover -race -timeout 10s ./pkg/... ./plugins/...
 
 integration-test: $(SOURCE)
 	go test -v -tags integration -race -timeout 10s ./pkg/... ./plugins/...
 
+e2e-test:  $(SOURCE) plugins
+	sudo -E CNI_PATH=${ROOT}/bin/plugins ${GO_EXECUTABLE} test -v -tags e2e -race -timeout 60s ./plugins/...
```
nmeyerhans: Since we rely on PATH for finding the go executable anyway, what's the advantage to `${GO_EXECUTABLE}` here instead of simply `go`?

samuelkarp: `PATH` for `sudo` is not necessarily the same as `PATH` for the user running this script (in my case, `go` is not in my global `PATH` but instead lives in `~/.go/bin`).

aaithal: This was the weirdness that I was running into:

```
[ec2-user@ip-1xx ~]$ go version
go version go1.7.6 linux/amd64
[ec2-user@ip-1xx ~]$ echo $PATH
/home/ec2-user/go176/go/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/opt/aws/bin:/home/ec2-user/.local/bin:/home/ec2-user/bin
[ec2-user@ip-1xx ~]$ sudo -E go version
sudo: go: command not found
[ec2-user@ip-1xx ~]$ sudo -E echo $PATH
/home/ec2-user/go176/go/bin:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/opt/aws/bin:/home/ec2-user/.local/bin:/home/ec2-user/bin
```

@nmeyerhans  Any suggestions for alternatives here which can get rid of the need for `${GO_EXECUTABLE}`?

Makefile
aaithal: done

Makefile
nmeyerhans: I think (at least on Amazon Linux) that we've got a value compiled in for the `secure_path` sudoers setting, and that's overriding PATH even when `-E` is in effect. So yeah, stick with `${GO_EXECUTABLE}`

aaithal: aha. makes sense.

PR #48, opened 2017-08-23
Add a parameter in ENI plugin to block Instance Metadata Service access

aaithal:
### Details
ENI plugin adds a `blackhole` route to `169.254.169.254` if `block-instance-metadata` config  parameter is set to `true`

From [README](https://github.com/aaithal/amazon-ecs-cni-plugins-1/blob/06af3daac1437a016e62ca959cb31a3acce81014/plugins/eni/README.md):
```
{
    "type":"ecs-eni",
    "cniVersion":"0.3.0",
    "eni":"eni-eni01en1",
    "ipv4-address":"172.31.31.65/20",
    "mac":"01:23:45:67:89:ab",
    "block-instance-metadata":true
}
```

### Testing Done
Verified that new updated tests pass:
* [X] unit tests
* [X] integration tests
* [X] end-to-end tests
* [X] manual tests: Manually verified that a container started using this plugin cannot access instance metadata service
```
$ ip netns exec 314e45c385dd ip route
default via xxx dev eth1
...
blackhole 169.254.169.254
169.254.170.2 via 169.254.172.1 dev ecs-eth0

$ ip netns exec 314e45c385dd curl 169.254.169.254
curl: (7) Couldn't connect to server
```

### DIFF
--- a/plugins/eni/engine/nsclosure.go
+++ b/plugins/eni/engine/nsclosure.go
@@ -15,6 +15,7 @@ package engine
 
 import (
 	"net"
+	"syscall"
 	"time"
 
 	"github.com/aws/amazon-ecs-cni-plugins/pkg/netlinkwrapper"
@@ -22,6 +23,10 @@ import (
 	"github.com/containernetworking/cni/pkg/ns"
 	"github.com/pkg/errors"
 	"github.com/vishvananda/netlink"
+)
+
+const (
+	instanceMetadataEndpoint = "169.254.169.254/32"
 )
 
 var linkWithMACNotFoundError = errors.New("engine: device with mac address not found")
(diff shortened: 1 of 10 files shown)


plugins/eni/engine/nsclosure.go
```diff
@@ -143,6 +150,22 @@ func (closureContext *setupNamespaceClosureContext) run(_ ns.NetNS) error {
...
 
+	// Add a blackhole route for IMDS endpoint if required
+	if closureContext.blockIMDS {
+		_, imdsNetwork, err := net.ParseCIDR(instanceMetadataEndpoint)
+		if err != nil {
+			// This should never happen because we always expect
+			// 169.254.169.254/32 to be parsed without any errors
+			return errors.Wrapf(err, "setupNamespaceClosure engine: unable to parse instance metadata endpoint")
+		}
+		if err = closureContext.netLink.RouteAdd(&netlink.Route{
```
richardpen: Just want to understand here, is this like: `ip route add blackhole 169.254.169.254/32`? What's the difference with `ip rule add blackhole to 169.254.169.254/32`?

PR #52, opened 2017-09-11
Skip IPAM del when IPAM isn't specified in the configuration

richardpen:
This PR will suppress the error caused by calling the bridge plugin without specifying the `ipam` plugin. This kind of behavior happens in ecs-agent: aws/amazon-ecs-agent#976

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,48 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"reflect"
+)
+
+// ZeroOrNil checks if the passed in interface is empty
+func ZeroOrNil(obj interface{}) bool {
+	if obj == nil {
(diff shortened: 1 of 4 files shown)


plugins/ecs-bridge/commands/commands.go
```diff
@@ -112,6 +114,11 @@ func del(args *skel.CmdArgs, engine engine.Engine) error {
 		return err
 	}
 
+	if utils.ZeroOrNil(conf.IPAM) {
+		log.Infof("IPAM configuration not fund, skip DEL for IPAM")
```
aaithal: spelling: `found`

pkg/utils/utils.go
```diff
@@ -0,0 +1,41 @@
...
+	"reflect"
+)
+
+// ZeroOrNil checks if the passed in interface is empty
+func ZeroOrNil(obj interface{}) bool {
+	value := reflect.ValueOf(obj)
+	if !value.IsValid() {
+		return true
+	}
+	if obj == nil {
```
aaithal: Shouldn't this be the first thing you check?

pkg/utils/utils.go
```diff
@@ -0,0 +1,41 @@
...
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"reflect"
+)
+
+// ZeroOrNil checks if the passed in interface is empty
+func ZeroOrNil(obj interface{}) bool {
```
aaithal: Please add unit tests for this.

pkg/utils/utils.go
```diff
@@ -0,0 +1,41 @@
...
+		return true
+	}
+	if obj == nil {
+		return true
+	}
+	switch value.Kind() {
+	case reflect.Slice, reflect.Array, reflect.Map:
+		return value.Len() == 0
+	}
+	zero := reflect.Zero(reflect.TypeOf(obj))
```
aaithal: Can you move this to after the `if` block, where it's actually used?

PR #54, opened 2017-10-02
IPv6 fixes

aaithal:
### Description
Benchmarking experiments show that there's usually 3-6s delay in
IPv6 address path being populated in Instance Metadata for ENI.

We also need to take into account delays in cleaning up of the
dhclient processes running in the default network namespace.

This commit adds retries for both of these to account for these
delays. This will result in a latency increase in surfacing errors
when there are failures such as wrong MAC Address being specified
in the CNI config. However, we err on the side of correctness,
fault tolerance and caution, and deem that as an acceptable cost.

### Testing Done
- [X] Build and Tests
```
$ make clean plugins unit-test integration-test e2e-test; echo $?
0
```
- [X] Manual Testing: Ran the plugin manually and verified that the container was configured with IPv6 IP address

### DIFF
--- a/plugins/eni/engine/dhclient.go
+++ b/plugins/eni/engine/dhclient.go
@@ -59,6 +59,9 @@ const (
 	// package and is not exposed anywhere. Hence, we declare this string to
 	// match the error message returned by the os package
 	processFinishedErrorMessage = "os: process already finished"
+
+	maxDHClientStartAttempts            = 10
+	durationBetweenDHClientStartRetries = 2 * time.Second
 )
 
 // DHClient interface provides methods to perform various dhclient actions
@@ -75,12 +78,14 @@ type DHClient interface {
 }
 
 type dhclient struct {
-	os             oswrapper.OS
-	exec           execwrapper.Exec
-	ioutil         ioutilwrapper.IOUtil
-	executable     string
-	leasesFilePath string
-	pidFilePath    string
+	os                                  oswrapper.OS
+	exec                                execwrapper.Exec
+	ioutil                              ioutilwrapper.IOUtil
+	executable                          string
+	leasesFilePath                      string
+	pidFilePath                         string
+	dhclientStartAttemptsMax            int
+	dhclientStartDurationBetweenRetries time.Duration
 }
 
 // NewDHClient creates a new DHClient object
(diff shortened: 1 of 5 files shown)


plugins/eni/engine/dhclient.go
```diff
@@ -131,14 +144,29 @@ func (client *dhclient) IsExecutableInPath() bool {
...
+	var err error
+	attempts := 1
+	for {
+		cmd := client.exec.Command(client.executable, args...)
+		out, err = cmd.CombinedOutput()
+		if err == nil {
+			return nil
+		}
+
+		log.Errorf("Error executing '%s' with args '%v' (attempt %d/%d): raw output: %s",
```
samuelkarp: Should this be an ERROR or a WARN?

plugins/eni/engine/engine.go
```diff
@@ -304,9 +317,28 @@ func (engine *engine) DoesMACAddressMapToIPV6Address(macAddress string, ipv6Addr
 
 func (engine *engine) doesMACAddressMapToIPAddress(macAddress string, addressToFind string, metatdataPathSuffix string) (bool, error) {
 	// TODO Use fmt.Sprintf and wrap that in a method
-	addressesResponse, err := engine.metadata.GetMetadata(metadataNetworkInterfacesPath + macAddress + metatdataPathSuffix)
+	var addressesResponse string
+	var err error
+
+	attempts := 1
+	for {
```
samuelkarp: You're just checking for `err != nil` outside this loop, so why not just make this a for loop with the count in the loop instead of the body?

```go
for attempts := 0; attempts < engine.metadataMaxRetryCount; attempts++
```

plugins/eni/engine/dhclient.go
```diff
@@ -131,14 +144,29 @@ func (client *dhclient) IsExecutableInPath() bool {
...
-	if err != nil {
-		log.Errorf("Error executing '%s' with args '%v': raw output: %s",
-			client.executable, args, string(out))
-		return errors.Wrapf(err,
-			"engine dhclient: unable to start dhclient for ipv%d address; command: %s %v; output: %s",
-			ipRev, client.executable, args, string(out))
+	var out []byte
+	var err error
+	attempts := 1
+	for {
```
samuelkarp: You're doing the same kind of retry in two different ways.  Maybe make them the same?  In particular:

* You can use the attempt count in the for loop itself `for attempts := 0; attempts < client.dhclientStartAttepmtsMax; attempts++`
* You can check for `err != nil` outside the loop and use that to log or return an error from the function

plugins/eni/engine/dhclient.go
```diff
@@ -59,6 +59,9 @@ const (
 	// package and is not exposed anywhere. Hence, we declare this string to
 	// match the error message returned by the os package
 	processFinishedErrorMessage = "os: process already finished"
+
+	maxDHClientStartAttempts            = 10
+	durationBetweenDHClientStartRetries = 2 * time.Second
```
richardpen: Do we have some data for this, I'm a little worried about 20 seconds is not enough?

aaithal: Yeah, in my tests, i've seen p100 at `5s`.

plugins/eni/engine/engine.go
aaithal: The reason for doing it this way is that I don't want an additional sleep on the last attempt if there's an error.  `for attempts := 0; attempts < engine.metadataMaxRetryCount; attempts++` doesn't give me that. Open to suggestions here.

plugins/eni/engine/dhclient.go
aaithal: `Warn` would be better. I'll modify that.

plugins/eni/engine/engine.go
samuelkarp: Got it, that makes sense.

PR #55, opened 2017-10-11
eni: ignore 'file exists' for default ipv6 route

aaithal:
Since the default ipv6 gateway can also be configured based on the
ipv6 address when the address is assigned, we sometimes get the
"file exists" error when adding the default route for the ipv6
gateway. This change ignores that error.

There are also logging improvements to log either the eni id or the
device name in various log messages for the eni plugin.

- [X] Build and Tests
```
$ make clean plugins unit-test integration-test e2e-test; echo $?
0
```
- [X] Manual Testing: Task start errors on an Instance with ipv6 enabled reduced to 0% from 40-60% with the modified eni plugin

### DIFF
--- a/plugins/eni/commands/commands.go
+++ b/plugins/eni/commands/commands.go
@@ -66,23 +66,24 @@ func add(args *skel.CmdArgs, engine engine.Engine, dhclient engine.DHClient) err
 	if err != nil {
 		return err
 	}
+	log.Infof("Found ENI with mac address on the host (id=%s): %s", conf.ENIID, macAddressOfENI)
 
 	// Get the interface name of the device by scanning links
 	networkDeviceName, err := engine.GetInterfaceDeviceName(macAddressOfENI)
 	if err != nil {
-		log.Errorf("Unable to find network device for the ENI: %v", err)
+		log.Errorf("Unable to find network device for ENI (mac address=%s): %v", macAddressOfENI, err)
 		return err
 	}
-	log.Debugf("Found network device for the ENI: %v", networkDeviceName)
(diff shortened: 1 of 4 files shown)


plugins/eni/commands/commands.go
```diff
@@ -66,23 +66,24 @@ func add(args *skel.CmdArgs, engine engine.Engine, dhclient engine.DHClient) err
 	if err != nil {
 		return err
 	}
+	log.Infof("Found ENI with mac address (id=%s): %s", conf.ENIID, macAddressOfENI)
```
richardpen: This message is very similar to the one below, can you add context eg: `from metedata` or `on the host` separately into those logs?

plugins/eni/commands/commands.go
```diff
@@ -187,20 +188,22 @@ func getMACAddressOfENI(conf *types.NetConf, engine engine.Engine) (string, erro
...
 	ok, err := engine.DoesMACAddressMapToIPV4Address(macAddressOfENI, ipv4Address)
 	if err != nil {
-		log.Errorf("Error validating ipv4 addresses for ENI: %v", err)
+		log.Errorf("Error validating ipv4 addresses for ENI (mac address=%s,ipv4 address=%s): %v",
+			macAddressOfENI, ipv4Address, err)
 		return err
 	}
 	if !ok {
-		log.Error("Unable to validate ipv4 address for ENI: %v", unmappedIPV4AddressError)
+		log.Error("Unable to validate ipv4 address for ENI (mac address=%s,ipv4 address=%s): %v",
```
richardpen: `Errorf`

PR #59, opened 2017-11-22
Respect CNI_IFNAME  passed in

lilida:
Summary:
This patch makes the plugin respect to the CNI_IFNAME. It sets the interface name and also redefine s the name schema of dhcp lease file to use mac address to avoid conflict on the lease file name

Tests:
Unit-tests, integration tests and e2e tests all passed. 
Tested in Kubernetes 1.8 cluster


aaithal:
@lilida thanks a lot for creating this pull request and apologies for the delayed response. I took a brief look at the changes and I don't see any major issues with it. I do however want to confirm that the change doesn't lead to any weird behavior in the ECS agent or when running tasks in ECS in `awsvpc` mode. I will get back to you when I have gathered that data. In the mean time, can you please confirm this contribution is under the terms of the Apache 2.0 license? 

Thanks,
Anirudh

lilida:
Thanks @aaithal  Totally understand and looking forward to getting it merged into the mainstream

lilida:
Forget to add. Yes, I confirm it is fine to release the code in Apache 2.0 license

pkg/netlinkwrapper/netlink.go
```diff
@@ -31,6 +31,8 @@ type NetLink interface {
 	LinkAdd(link netlink.Link) error
 	// LinkSetMaster is equivalent to `ip link set $link master $master`
 	LinkSetMaster(link netlink.Link, master *netlink.Bridge) error
+	//LinkSetName sets the interface name
```
aaithal: nit: please leave a space after `// `

plugins/eni/engine/dhclient.go
```diff
@@ -185,25 +185,25 @@ func (client *dhclient) constructDHClientArgs(deviceName string, ipRev int) []st
 	}
 	return append(commonArgs,
 		"-lf",
-		client.getLeasesFile(deviceName, ipRev), // The path to the leases file
+		client.getLeasesFile(macAddress, ipRev), // The path to the leases file
```
aaithal: what is the rationale for changing this from device name to mac address?

pkg/netlinkwrapper/netlink.go
lilida: Done

plugins/eni/engine/dhclient.go
lilida: In Kubernetes, CNI passed in eth0 for the CNI_IFNAME. The device is always renamed to eth0 for all Pods and we cannot create the lease file with name in the container network namespace. We also cannot use the host interface name. This is because the name is changing. The name depends on how many network interfaces in host network namespace at the moment. One interface can be eth1 on the host, the be renamed to eth0 in container network namespace and then be changed to eth2 if eth0 and eth1 are there when the pod is destroyed. Therefore, it is preferrable to have a unique identity for each network interface. MacAddress is in use with this context.

aaithal: Thanks. My question was more to the point that today, the lease file is named something like `ns-eth2-dhclient4.leases`. With this change, it seems like, we'll be using the mac address instead of `eth2` IIUC (`ns-aa:bb:..-dhclient4.leases`). Please correct me if i'm wrong.

lilida: Yes. You are right

aaithal: Thanks for clarifying. I understand it better now.

PR #62, opened 2017-12-26
Adding retry for eni plugin to acquire mac address

richardpen:
This PR adds backoff retry logic for retrieving mac address of the eni from ec2 instance metadata.

### DIFF
--- a/pkg/utils/backoff.go
+++ b/pkg/utils/backoff.go
@@ -1,0 +1,83 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"math"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+// Backoff defines the interface for a backoff
+type Backoff interface {
+	Reset()
+	Duration() time.Duration
+}
+
+// SimpleBackoff implements a simple mechanism for backoff
+type SimpleBackoff struct {
+	current        time.Duration
+	start          time.Duration
+	max            time.Duration
+	jitterMultiple float64
+	multiple       float64
+	mu             sync.Mutex
+}
+
+// NewSimpleBackoff creates a Backoff which ranges from min to max increasing by
+// multiple each time.
+// It also adds (and yes, the jitter is always added, never
+// subtracted) a random amount of jitter up to jitterMultiple percent (that is,
+// jitterMultiple = 0.0 is no jitter, 0.15 is 15% added jitter). The total time
+// may exceed "max" when accounting for jitter, such that the absolute max is
+// max + max * jiterMultiple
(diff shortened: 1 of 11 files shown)


plugins/eni/commands/commands.go
```diff
@@ -152,6 +163,63 @@ func add(args *skel.CmdArgs, engine engine.Engine, dhclient engine.DHClient) err
...
+	defer cancel()
+
+	err := utils.RetryWithBackoffCtx(ctx, backoff, func() error {
+		mac, err := getMACAddressOfENI(retriever.conf, retriever.engine)
+		if err == nil {
+			retriever.macAddress = mac
+			return nil
+		}
+
+		_, ok := err.(*engine.UnmappedMACAddressError)
```
nmeyerhans: I realize that we don't actually have the issue here, but this style of type assertion gives me flashbacks to https://github.com/aws/amazon-ecs-agent/pull/910. [Interface conversions](https://golang.org/doc/effective_go.html#interface_conversions) are a less error prone way to perform type assertions, IMO.

plugins/eni/commands/commands.go
```diff
@@ -27,6 +30,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	ec2InstanceMetadataBackoffMin      = 500 * time.Millisecond
+	ec2InstanceMetadataBackoffMax      = 1 * time.Second
+	ec2InstanceMetadataBackoffMultiple = 2
+	ec2InstanceMetadataBackoffJitter   = 0.2
+	ec2InstanceMetadataTimeout         = 5 * time.Second
```
nmeyerhans: How did you arrive at a 5 second timeout?

pkg/utils/backoff.go
```diff
@@ -0,0 +1,79 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
```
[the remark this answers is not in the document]
richardpen: yeah, I don't like copy and paste either, but it will require additional work which probably won't happen within this pr, I will create an issue to follow up.

plugins/eni/commands/commands.go
richardpen: @aaithal  and I did some tests, the worst case  I saw is 1.5s. So I put 5s here to be safe, but it's changeable.

pkg/utils/utils.go
```diff
@@ -46,3 +48,36 @@ func ZeroOrNil(obj interface{}) bool {
...
+// If the context is done, nil will be returned
+// If the error is nil then the function will no longer be called
+// If the error is Retriable then that will be used to determine if it should be
+// retried
+func RetryWithBackoffCtx(ctx context.Context, backoff Backoff, fn func() error) error {
+	var err error
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
aaithal: nice. we need to fix this in the agent repo as well.

plugins/eni/commands/commands.go
```diff
@@ -27,6 +30,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	ec2InstanceMetadataBackoffMin      = 500 * time.Millisecond
```
aaithal: can you decrease this to `100ms`? half a second seems really long to wait here.

plugins/eni/commands/commands.go
aaithal: Can you address @nmeyerhans 's comment here as well?

PR #67, opened 2018-01-11
update the version to 2018.01.0

richardpen:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
+## 2017.10.2
```
aaithal: this should be 2018.01.0

PR #73, opened 2018-02-23
Add a route for arp query request from host

richardpen:
This PR is trying to fix issue: aws/amazon-ecs-agent#1230 by adding a route for the arp query from host,
After adding this, the route inside the container will look like:
```
root@797087bca637:/# ip route
default via 172.31.192.1 dev eth0
169.254.170.2 via 169.254.172.1 dev ecs-eth0
169.254.172.1 via 169.254.172.1 dev ecs-eth0
172.31.192.0/20 dev eth0  proto kernel  scope link  src 172.31.198.99
```

The issue can be reproduced without this change but not with this change  by the following steps:
* inside container: curl 169.254.170.2/v2/metadata
* inside container: arp -a
* host: sudo arp -d 169.254.172.2
* inside container: curl  --connection-timeout 5 169.254.170.2/v2/metadata  (this will timeout without the change)

### DIFF
--- a/plugins/ecs-bridge/e2eTests/e2e_test.go
+++ b/plugins/ecs-bridge/e2eTests/e2e_test.go
@@ -39,6 +39,7 @@ const (
 	expectedGateway       = "169.254.172.1"
 	expectedVethAddress   = "169.254.172.2/22"
 	dst                   = "169.254.170.2/32"
+	bridgeDst             = "169.254.172.1/32"
 	netConf               = `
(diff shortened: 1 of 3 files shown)


ofiliz:
The route table in the description looks weird. APIPA range is link-local by definition, and there shouldn't be a need to set a route to a specific neighbor in the same link-local subnet. If there is no urgency, can we discuss this before merging?

PR #81, opened 2018-05-30
eni: add ipv4 subnet gateway from config

aaithal:
### Summary

Add functionality to parse ipv4 subnet gateway from config. This eliminates the dependency on instance metadata service for the ENI plugin, which should hopefully reduce the ENI setup time.

### Implementation details

Added a new config field for supplying ipv4 subnet gateway address for
the caller. When that's specified the plugin can use that directly
instead of looking up that information from instance metadata service.

Also, got rid of the check where the plugin makes sure that the eni id
and the mac address match by looking up instance metadata. This leads to
addtional delays starting the plugin while adding very little value to
the functionality. Hence, the ENI plugin doesn't do that check anymore.
It's left to the caller instead should make sure that correct values
are passed to the plugin, which the ECS agent is currently doing anyway.

### Testing

New tests cover the changes: Yes

| Description | End-to-end test duration for setting up the ENI (Average of 5 runs) | 
|------------ |:-----------------------------------------------:|
| With instance metadata interaction | 1278 ms |
| Without instance metadata interaction | 142.5 ms |

Added a changelog commit

### Licensing

This contribution is under the terms of the Apache 2.0 License: yes

### DIFF
--- a/pkg/utils/gateway.go
+++ b/pkg/utils/gateway.go
@@ -1,0 +1,116 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	minIPV4CIDRBlockSize = 28
+	maxIPV4CIDRBlockSize = 16
+)
+
+// ParseIPV4GatewayNetmaskError is used to indicate any error with parsing the
+// IPV4 address and the netmask of the ENI
+type ParseIPV4GatewayNetmaskError struct {
+	operation string
+	origin    string
+	message   string
+}
+
+func (err *ParseIPV4GatewayNetmaskError) Error() string {
+	return err.operation + " " + err.origin + ": " + err.message
+}
+
+func newParseIPV4GatewayNetmaskError(operation string, origin string, message string) error {
+	return &ParseIPV4GatewayNetmaskError{
+		operation: operation,
+		origin:    origin,
+		message:   message,
+	}
+}
+
+// ComputeIPV4GatewayNetmask computes the subnet gateway and netmask for
+// the gateway given the ipv4 cidr block for the ENI
+// Gateways are provided in VPC subnets at base +1
+// (NOTE: from the base of the subnet, not the VPC base)
+// https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html#vpc-sizing-ipv4
+func ComputeIPV4GatewayNetmask(cidrBlock string) (string, string, error) {
+	// The IPV4 CIDR block is of the format ip-addr/netmask
+	ip, ipNet, err := net.ParseCIDR(cidrBlock)
+	if err != nil {
+		return "", "", errors.Wrapf(err,
+			"compute ipv4 gateway netmask: unable to parse cidr: '%s'", cidrBlock)
+	}
+
+	ip4 := ip.To4()
+	if ip4 == nil {
+		return "", "", newParseIPV4GatewayNetmaskError("compute ipv4 gateway netmask", "engine",
+			fmt.Sprintf("unable to parse ipv4 gateway from cidr block '%s'", cidrBlock))
+	}
+
+	maskOnes, _ := ipNet.Mask.Size()
+	// As per
+	// http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html#VPC_Sizing
+	// You can assign a single CIDR block to a VPC. The allowed block size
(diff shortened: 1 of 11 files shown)


pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+	ip, ipNet, err := net.ParseCIDR(cidrBlock)
+	if err != nil {
+		return "", "", errors.Wrapf(err,
+			"compute ipv4 gateway netmask: unable to parse cidr: '%s'", cidrBlock)
+	}
+
+	ip4 := ip.To4()
+	if ip4 == nil {
+		return "", "", newParseIPV4GatewayNetmaskError("compute ipv4 gateway netmask", "engine",
+			fmt.Sprintf("unable to parse ipv4 gateway from cidr block '%s'", cidrBlock))
```
richardpen: Can you change it to `errors.Errorf("xxx %q",cirdrBlock)`?

pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+	maskOnes, _ := ipNet.Mask.Size()
+	// As per
+	// http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html#VPC_Sizing
+	// You can assign a single CIDR block to a VPC. The allowed block size
+	// is between a /16 netmask and /28 netmask. Verify that
+	if maskOnes > minIPV4CIDRBlockSize {
+		return "", "", errors.Errorf("compute ipv4 gateway netmask: invalid ipv4 cidr block, %d > 28", maskOnes)
+	}
+	if maskOnes < maxIPV4CIDRBlockSize {
+		return "", "", errors.Errorf("compute ipv4 gateway netmask: invalid ipv4 cidr block, %d <= 16", maskOnes)
```
richardpen: nit: `%d < 16`.

pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+	operation string
+	origin    string
+	message   string
+}
+
+func (err *ParseIPV4GatewayNetmaskError) Error() string {
+	return err.operation + " " + err.origin + ": " + err.message
+}
+
+func newParseIPV4GatewayNetmaskError(operation string, origin string, message string) error {
```
richardpen: I think this should be public so that other packages can reuse this error?

pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+	// is between a /16 netmask and /28 netmask. Verify that
+	if maskOnes > minIPV4CIDRBlockSize {
+		return "", "", errors.Errorf("compute ipv4 gateway netmask: invalid ipv4 cidr block, %d > 28", maskOnes)
+	}
+	if maskOnes < maxIPV4CIDRBlockSize {
+		return "", "", errors.Errorf("compute ipv4 gateway netmask: invalid ipv4 cidr block, %d <= 16", maskOnes)
+	}
+
+	// ipv4 gateway is the first available IP address in the subnet
+	ip4[3] = ip4[3] + 1
```
richardpen: nit: `ip4[3] ++` is simpler?

pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// ParseIPV4GatewayNetmask parses the cidr block to return the subnet gateway
+// ip and the netmask. There's no additional computation here. It essentially
+// splits the string on '/' and returns the parts. However, it uses golang's
+// 'net' package to do this. It also performs additional validation of the
+// cidr block, which is typically passed to the plugin as a config
+func ParseIPV4GatewayNetmask(cidrBlock string) (string, string, error) {
+	// The IPV4 CIDR block is of the format ip-addr/netmask
+	ip, ipNet, err := net.ParseCIDR(cidrBlock)
```
richardpen: These validations are duplicate code, can you wrapper them into a single method?

pkg/utils/gateway.go
aaithal: This is not being used externally today. We can make it public if/when needed.

pkg/utils/gateway.go
aaithal: They have differing error messages. Doesn't seem worth it to create a wrapper method for just that.

plugins/eni/engine/engine_test.go
```diff
@@ -18,12 +18,13 @@ package engine
 import (
 	"errors"
 	"fmt"
-	"github.com/containernetworking/cni/pkg/skel"
 	"net"
 	"syscall"
 	"testing"
 	"time"
 
+	"github.com/containernetworking/cni/pkg/skel"
```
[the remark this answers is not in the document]
aaithal: I'm not sure. This was the result from `gofmt`

pkg/utils/gateway.go
```diff
@@ -0,0 +1,116 @@
...
+		return "", "", newParseIPV4GatewayNetmaskError("compute ipv4 gateway netmask", "engine",
+			fmt.Sprintf("unable to parse ipv4 gateway from cidr block '%s'", cidrBlock))
+	}
+
+	maskOnes, _ := ipNet.Mask.Size()
+	// As per
+	// http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html#VPC_Sizing
+	// You can assign a single CIDR block to a VPC. The allowed block size
+	// is between a /16 netmask and /28 netmask. Verify that
+	if maskOnes > minIPV4CIDRBlockSize {
```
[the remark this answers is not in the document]
aaithal: I agree that we can't prevent the caller from passing incorrect subnet with the correct length. We have in the past seen incorrect payloads being sent from ECS backend because of bugs etc. It's better to fail early in those scenarios than have the applications fail with network erorrs in those scenarios

PR #83, opened 2018-08-16
Remove dhclient from eni plugin

richardpen:
### Testing

unit-test passed
manual test passed

New tests cover the changes: 

### Licensing

This contribution is under the terms of the Apache 2.0 License:


ofiliz:
Instead of removing it, we should make the dhclient configurable.

PR #86, opened 2018-10-09
update the version to 2018.10.0 in master

adnxn:
### Summary
Merging dev into master 

### Testing

New tests cover the changes: N/A

### Licensing


yunhee-l:
this is just merging dev into master, correct? can you add some details in summary at least?

adnxn:
> this is just merging dev into master, correct? can you add some details in summary at least?

will do.

sharanyad:
not sure if this is the right place, but adding a note here. we should update [this](https://github.com/aws/amazon-ecs-agent/pull/1592/) test when we update cni-plugins in agent.

adnxn:
> not sure if this is the right place, but adding a note here. we should update this test when we update cni-plugins in agent.

thanks for calling it out. i think moving forward we should be updating that test when we pull a new version of the plugin into the agent repo.

PR #90, opened 2019-06-03
Fix race condition when create/configure the bridge

fenxiong:
### Summary

Due to the fact that multiple tasks may be modifying the bridge at the same time, there are race conditions when creating/configuring the bridge, which cause the following error:
```
# create bridge race
unable to configure pause container namespace: cni setup: invoke bridge plugin failed: bridge create: unable to add bridge interface ecs-bridge: file exists

# configure bridge race
unable to configure pause container namespace: cni setup: invoke bridge plugin failed: bridge configure: unable to assign ip address to bridge ecs-bridge: file exists
```
This PR fixes the two race conditions.

### Implementation details

When creating the bridge, check whether the bridge exists after failing to create it.
When assigning ip address to the bridge, check whether the ip address is already assigned after failing to assign it.

### Testing

Unit tests added. Following targets passed:
make unit-test
make integration-test
make e2e-test
make plugins

Manually tested by building this into an agent and ran awsvpc tasks, verified that race condition failure not happening with the fix.

New tests cover the changes:  yes

### Licensing

### DIFF
--- a/plugins/ecs-bridge/engine/engine.go
+++ b/plugins/ecs-bridge/engine/engine.go
@@ -15,6 +15,7 @@ package engine
 
 import (
 	"net"
+	"strings"
 	"syscall"
 
 	"github.com/aws/amazon-ecs-cni-plugins/pkg/cniipamwrapper"
@@ -26,10 +27,14 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-// zeroLengthIPString is what we expect net.IP.String() to return if the
-// ip has length 0. We use this to determing if an IP is empty.
-// Refer https://golang.org/pkg/net/#IP.String
-const zeroLengthIPString = "<nil>"
+const  (
+	// zeroLengthIPString is what we expect net.IP.String() to return if the
+	// ip has length 0. We use this to determing if an IP is empty.
+	// Refer https://golang.org/pkg/net/#IP.String
+	zeroLengthIPString = "<nil>"
+
+	fileExistsErrMsg = "file exists"
+)
 
 // Engine represents the execution engine for the ECS Bridge plugin.
 // It defines all the operations performed during the execution of the
@@ -73,7 +78,12 @@ func (engine *engine) CreateBridge(bridgeName string, mtu int) (*netlink.Bridge,
 	if bridge == nil {
 		err = engine.createBridge(bridgeName, mtu)
 		if err != nil {
-			return nil, err
+			if !strings.Contains(err.Error(), fileExistsErrMsg) {
+				return nil, err
+			}
+			// If the error returned by createBridge is that the bridge already exists, proceed to
+			// lookupBridge because that means the bridge was created by someone else right before
+			// we tried creating it, which is fine
 		}
 
 		// We need to lookup the bridge link again because LinkAdd
@@ -241,14 +251,21 @@ func (engine *engine) ConfigureBridge(result *current.Result, bridge *netlink.Br
 	bridgeAddr := &netlink.Addr{
 		IPNet: resultBridgeNetwork,
 	}
-	err = engine.netLink.AddrAdd(bridge, bridgeAddr)
-	if err != nil {
-		return errors.Wrapf(err,
-			"bridge configure: unable to assign ip address to bridge %s",
-			bridge.Attrs().Name)
+
+	addrAddErr := engine.netLink.AddrAdd(bridge, bridgeAddr)
+	if addrAddErr == nil {
+		return nil
 	}
 
-	return nil
+	if strings.Contains(addrAddErr.Error(), fileExistsErrMsg) {
+		// if we fail to assign ip address because of a file exist error, the error is probably caused by someone else
+		// assigned the address right before we assigned it, which is fine
+		return nil
+	}
+
+	return errors.Wrapf(addrAddErr,
+		"bridge configure: unable to assign ip address to bridge %s",
+		bridge.Attrs().Name)
 }
 
 // GetInterfaceIPV4Address gets the ipv4 address of a given interface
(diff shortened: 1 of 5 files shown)


plugins/ecs-bridge/engine/engine.go
```diff
@@ -73,7 +74,9 @@ func (engine *engine) CreateBridge(bridgeName string, mtu int) (*netlink.Bridge,
 	if bridge == nil {
 		err = engine.createBridge(bridgeName, mtu)
 		if err != nil {
-			return nil, err
+			// If createBridge returns error, still proceed to lookupBridge because the error
+			// may just indicate someone else already created the bridge
+			log.Warnf("Unable to create bridge: %v", err)
```
aaithal: I think a better place to log this error is in `commands.go`, [here](https://github.com/aws/amazon-ecs-cni-plugins/blob/0c6216c60401232805e50e31d4040ae84b0b23cf/plugins/ecs-bridge/commands/commands.go#L30). That way, we don't have to add a `log` statement every time an error is returned and instead log it in one place (actually fix the `TODO` [here](https://github.com/aws/amazon-ecs-cni-plugins/blob/0c6216c60401232805e50e31d4040ae84b0b23cf/plugins/ecs-bridge/commands/commands.go#L42). We can add additional context by wrapping this with `errors.Wrap()`. At the very least, we should be logging this in `commands.go`, [where it's being invoked](https://github.com/aws/amazon-ecs-cni-plugins/blob/0c6216c60401232805e50e31d4040ae84b0b23cf/plugins/ecs-bridge/commands/commands.go#L49).

plugins/ecs-bridge/engine/engine.go
```diff
@@ -217,38 +220,59 @@ func (engine *engine) ConfigureContainerVethInterface(netnsName string, result *
...
+	if exists {
+		return nil
+	}
+
+	return errors.Wrapf(err,
+		"bridge configure: unable to assign ip address to bridge %s",
+		bridge.Attrs().Name)
+}
+
+func (engine *engine) bridgeNetworkExists(bridge *netlink.Bridge, bridgeNetworkCIDR string) (bool, error) {
```
aaithal: The way this method is written, you're always returning `nil` error value with `true` and a `non-nil` `error` value with `false`. Since you're ignoring the `error` returned anyway, you might as well not return it and just make this function return a `bool`. You can log the errors instead of returning and ignoring them to provide the visibility in log files.

plugins/ecs-bridge/engine/engine.go
```diff
@@ -217,38 +220,59 @@ func (engine *engine) ConfigureContainerVethInterface(netnsName string, result *
...
+	}
+
+	// if we fail to assign ip address, check bridge network exists or not again, because the error may be caused by
+	// someone else assigned the address right before we assigned it
+	exists, _ = engine.bridgeNetworkExists(bridge, resultBridgeCIDR)
+	if exists {
+		return nil
+	}
+
+	return errors.Wrapf(err,
```
aaithal: This is potentially a bug since you've already dealt with error in line 242. Please move this up to line 242 as a `if err != nil`  block.

plugins/ecs-bridge/engine/engine.go
fenxiong: i added the log in commands.go and fix the TODO. i think wrapping the createBridge error together with the error from lookupBridge seems weird though, so still keeping the warning log here..

plugins/ecs-bridge/engine/engine.go
fenxiong: to fix the race condition, we need to proceed to bridgeNetworkExists even if AddrAdd returns error in line 242, so i don't think we can return error after AddrAdd..

plugins/ecs-bridge/engine/engine.go
fenxiong: the function can return both (false, nil) and (false, err). the error from line 229 is not ignored which keeps the existing behavior before assigning ip address. the change is to add another address check after assigning the address

plugins/ecs-bridge/engine/engine.go
aaithal: Can you use a different variable name (than `err`) to store the return value in line 241? If someone adds any other logic in between lines 242 and 253 in the future and uses `err` to store the return value, this logic will become buggy.

fenxiong: updated.

plugins/ecs-bridge/engine/engine.go
aaithal: Right, but you're ignoring the error returned anyway. So, that information is being lost. It's better to at least log it within the method and make it return just `bool` instead.

fenxiong: ok, updated

plugins/ecs-bridge/engine/engine.go
```diff
@@ -217,38 +220,54 @@ func (engine *engine) ConfigureContainerVethInterface(netnsName string, result *
...
-			bridge.Attrs().Name)
+	addrAddErr := engine.netLink.AddrAdd(bridge, bridgeAddr)
+	if addrAddErr == nil {
+		return nil
 	}
 
-	return nil
+	// if we fail to assign ip address, check bridge network exists or not again, because the error may be caused by
+	// someone else assigned the address right before we assigned it
+	if engine.bridgeNetworkExists(bridge, resultBridgeCIDR) {
```
richardpen: This is assuming all the errors are caused by the "bridge already exists" which isn't true. If we ignore the other errors, the bridge may still be created by not working as expected. I think another way is to check the error message only for the conflict. But this should be covered by a integration test in case the error message from your dependency library changed.

PR #94, opened 2019-08-20
Stay down

aaithal:
### Summary

Adds `"stay-down"` config to ENI plugin.

The `stay-down` config specifies if the interface should stay down. If this is set, routes, addresses etc for the interface will not be set.

### Implementation details

* *eni plugin: don't bring up eni when 'stay-down' is set*:  Modified the eni plugins engine code to only move the ENI to the target network namespace when 'stay-down' flag is set to 'true'.

* *eni plugin: adding 'stay-down' config*:  The `stay-down` config specifies if the interface should stay down. If this is set, routes, addresses etc for the interface will not be set.

### Testing

Ran `make test integ-test ec2-test` without any errors.

New tests cover the changes:  yes

updated changelog with: 

```
* Enhancement - Add `stay-down` config to ENI plugin. [#94](https://github.com/aws/amazon-ecs-cni-plugins/pull/94)
```
### Licensing


plugins/eni/types/types_test.go
```diff
@@ -22,116 +22,86 @@ import (
...
-var configNoIPV4Address = `{"eni":"eni1", "mac":"01:23:45:67:89:ab"}`
-var configNoMAC = `{"eni":"eni1", "ipv4-address":"10.11.12.13"}`
-var configInvalidIPV4AddressMalformed = `{"eni":"eni1", "ipv4-address":"1", "mac":"01:23:45:67:89:ab"}`
-var configInvalidIPV4AddressIPV6 = `{"eni":"eni1", "ipv4-address":"2001:db8::68", "mac":"01:23:45:67:89:ab"}`
-var configMalformedMAC = `{"eni":"eni1", "ipv4-address":"10.11.12.13", "mac":"01:23:45:67:89"}`
-var validConfigWithIPV6 = `{
+var (
+	validConfigNoIPV6                 = `{"eni":"eni1", "ipv4-address":"10.11.12.13", "mac":"01:23:45:67:89:ab"}`
+	configNoENIID                     = `{"ipv4-address":"10.11.12.13", "mac":"01:23:45:67:89:ab"}`
+	configNoIPV4Address               = `{"eni":"eni1", "mac":"01:23:45:67:89:ab"}`
```
ofiliz: Is it still an invalid config if no IPv4 address is specified but stay-down is set to true? It feels like IPAddress fields should be optional when stay-down==true. I guess agent will never leave the IPv4 address field empty, so this is a pureness argument rather than an actual requirement, so I'm fine with the way the code is now.

plugins/eni/README.md
```diff
@@ -13,7 +13,8 @@ An example configuration for invoking the plugin is listed next:
     "eni":"eni-eni01en1",
     "ipv4-address":"172.31.31.65/20",
     "mac":"01:23:45:67:89:ab",
-    "block-instance-metadata":true
+    "block-instance-metadata":true,
+    "stay-down":true,
```
ofiliz: Nit: It might make more sense to have stay-down:false in the example, since that's what most tasks would do.

plugins/eni/types/types_test.go
aaithal: > Is it still an invalid config if no IPv4 address is specified but stay-down is set to true?

I think that config should be valid. But, I don't know if we need to relax these constraints yet. As you point out, these fields are being populated by the agent anyway. We can cross that bridge when we need to I guess.

plugins/eni/README.md
aaithal: Good point. done.

PR #104, opened 2023-01-19
Add -buildvcs=false to Makefile

SreeeS:
### Summary

Make release is failing on [amazon-ecs-agent](https://github.com/aws/amazon-ecs-agent) when we update the submodules.
 
The Error message is:
```
GOOS=linux CGO_ENABLED=0 go build -installsuffix cgo -a -ldflags "\
     -X github.com/aws/amazon-ecs-cni-plugins/pkg/version.GitShortHash=a04bd6ec \
     -X github.com/aws/amazon-ecs-cni-plugins/pkg/version.GitPorcelain=0 \
     -X github.com/aws/amazon-ecs-cni-plugins/pkg/version.Version=2020.09.0 -s" \
     -o /go/src/github.com/aws/amazon-ecs-cni-plugins/bin/plugins/ecs-eni github.com/aws/amazon-ecs-cni-plugins/plugins/eni
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:40: bin/plugins/ecs-eni] Error 1
make: *** [build-ecs-cni-plugins] Error 2
```
This change is to include `-buildvcs=false` flag to the go build in Make file.

This error occurred after we updated the amazon-ecs-cni-plugins sub module in amazon-ecs-agent to pull the latest commit due to this change in golang: https://github.com/golang/go/issues/37475

### Implementation details

Added `-buildvcs=false` flag in Make file.
`-buildvcs=false` excludes version control information when golang version is >= 1.18.

### Testing

make unit-test
make release on ecs-agent repo

New tests cover the changes:  no

bug-Fixed failed make release on ecs-agent repo

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ LOCAL_IPAM_PLUGIN_BINARY=bin/plugins/ecs-ipam
 LOCAL_BRIDGE_PLUGIN_BINARY=bin/plugins/ecs-bridge
 VERSION=$(shell cat $(ROOT)/VERSION)
 GO_EXECUTABLE=$(shell command -v go 2> /dev/null)
+GO_VERSION=`go version | cut -d' ' -f3 | cut -c 3-`
 
 # We want to embed some git details in the build. We support pulling
 # these details from the environment in order support builds outside
@@ -25,6 +26,15 @@ ifeq ($(strip $(GIT_PORCELAIN)),)
   GIT_PORCELAIN=1
 endif
 
+# buildvcs=false excludes version control information in golang >= 1.18. 
+# This is required for compiling agent with included repositories
+ifeq ($(shell expr $(GO_VERSION) \>= 1.18), 1)
+	BUILD_VCS := -buildvcs=false
+else
+	# leaving BUILD_VCS empty as it is not required for golang < 1.18.
+	BUILD_VCS :=
+endif
+
 .PHONY: get-deps
 
 get-deps:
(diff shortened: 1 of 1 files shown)


mythri-garaga:
Any implications of this setting to flag to `false`? Can we mention that in the summary too?

sparrc:
Would be good to mention in the summary why and when this started failing, I believe this is only applicable to when the plugins are used as git submodules, and due to this change in golang correct? https://github.com/golang/go/issues/37475

SreeeS:
> Any implications of this setting to flag to `false`? Can we mention that in the summary too?

Generally, setting the `buildvcs` to `false` will disable VCS stamping. Disabling the VCS stamping can make it harder to understand how it was built, hard to track and reproduce the issue in case of errors. In our case we are using this as a sub module plugin as this would not be in the root directory there seems to be a discrepancy in status of VCS, which is making the make release to fail on agent repo when we update the cni-plugin sub modules.

sparrc:
> Disabling the VCS stamping can make it harder to understand how it was built, hard to track and reproduce the issue in case of errors.

what does this mean? what specifically does this flag disable? Are there any features we currently use that this disables?

prateekchaudhry:
https://tip.golang.org/doc/go1.18

This flag seems to be introduced in golang 1.18. Will this change be backwards compatible with go version of the system? Do we want to make this flag conditional on go version?

SreeeS:
> Would be good to mention in the summary why and when this started failing, I believe this is only applicable to when the plugins are used as git submodules, and due to this change in golang correct? [golang/go#37475](https://github.com/golang/go/issues/37475)

I agree, update the summary.

SreeeS:
> > Disabling the VCS stamping can make it harder to understand how it was built, hard to track and reproduce the issue in case of errors.
> 
> what does this mean? what specifically does this flag disable? Are there any features we currently use that this disables?

I was mentioning general implications. We don't depend on it.

sparrc (approved):

SreeeS:
> https://tip.golang.org/doc/go1.18
> 
> This flag seems to be introduced in golang 1.18. Will this change be backwards compatible with go version of the system? Do we want to make this flag conditional on go version?

Added logic to use the flag based on golang version

prateekchaudhry (approved):

PR #106, opened 2023-02-15
Fix tests and DelLinkByNameAddr issue

yinyic:
### Summary

Fix unit, integ, and e2e tests, and an "index out of range" error from DelLinkByNameAddr

#### Unit test
Update `TestDelRunIPAMPluginDelError` from `plugins/ecs-bridge/commands/commands_test.go` since we updated code logic 

#### Integ test
`TestAssignReleaseExistGetAvailableIP` test is failing during `DEL` operation for
```
fatal error: checkptr: converted pointer straddles multiple allocations
```
And it leads to an issue in older version of boltdb (this is also mentioned [here](https://github.com/golang/go/issues/54690)), a dependency of `github.com/docker/libkv`. Upgrading `libkv` package resolves the issue since newer `libkv` uses newer boltdb.
```
go get github.com/docker/libkv
go mod tidy
go mod vendor
```

#### E2E test
* `containernetworking` repo removed some of their netns functionalities such as ns.NewNS() (it got moved to their test package). To workaround, this PR created `testutils` package and added two files (from the sister amazon-vpc-cni-plugins [repo](https://github.com/aws/amazon-vpc-cni-plugins/tree/master/network/netns) that contain useful functions to aid test resource management.
* CNI interface has been updated and requires more config parameters during invocation. The PR made those changes for the invocation to be successful.
* `DelLinkByNameAddr` in `pkg/cniipwrapper/ip.go` currently returns ips[0] when there's no error removing an interface. The `ips` list, however, could be empty even when error is nil. It turns out that the address of the interface (such as `169.254.172.2/22`) gets filtered out by upstream for its CIDR notation. We will now simply ignore the returned value, since we aren't making any use of it.
   * note that this issue may be observed in ECS Agent log in 1.68.1 release, with the error manifested as `panic: runtime error: index out of range [0] with length 0\\n\\ngoroutine 21 [running]:\\ngithub.com/aws/amazon-ecs-cni-plugins/pkg/cniipwrapper.(*cniIP).DelLinkByNameAddr...`

#### PR template
Updated repo PR template to add reminder for running make test targets.

### Testing

New tests cover the changes: 

#### All tests below should pass:
- [ ] make unit-test
- [ ] make integration-test
- [ ] make e2e-test

### Licensing


mythri-garaga (approved):

pkg/cniipwrapper/ip.go
```diff
@@ -47,11 +47,8 @@ func (*cniIP) SetHWAddrByIP(ifName string, ip4 net.IP, ip6 net.IP) error {
 	return ip.SetHWAddrByIP(ifName, ip4, ip6)
 }
 
-func (*cniIP) DelLinkByNameAddr(ifName string) (*net.IPNet, error) {
+func (*cniIP) DelLinkByNameAddr(ifName string) error {
 	// this was updated to return an array of its addresses.  For now just returning the first
```
sparrc, line 51: nit: this comment can be deleted

pkg/cniipwrapper/ip.go
```diff
@@ -47,11 +47,8 @@ func (*cniIP) SetHWAddrByIP(ifName string, ip4 net.IP, ip6 net.IP) error {
 	return ip.SetHWAddrByIP(ifName, ip4, ip6)
 }
 
-func (*cniIP) DelLinkByNameAddr(ifName string) (*net.IPNet, error) {
+func (*cniIP) DelLinkByNameAddr(ifName string) error {
```
sparrc, line 50: why is the ip return being deleted from the func signature? is it not relevant?

sparrc: nvm I see it in the description under e2e tests, another nitpick but maybe put this bit in the "summary" and "Description for changelog" since this is the main purpose of this PR, IIUC

pkg/cniipwrapper/ip.go
yinyic: Ah yes will remove it.

pkg/cniipwrapper/ip.go
yinyic: Yeah we don't use it. 
https://github.com/aws/amazon-ecs-cni-plugins/blob/a04bd6ecae16f6b2bc2957e3cafbf365c0710b19/plugins/ecs-bridge/engine/deletelink.go#L41

yinyic: > nvm I see it in the description under e2e tests, another nitpick but maybe put this bit in the "summary" and "Description for changelog" since this is the main purpose of this PR, IIUC

Np will do!

yinyic:
Closing PR - we've reverted dependency updates hence quite a lot of changes in this PR are no longer applicable.

PR #111, opened 2023-05-31
Set static MAC address when creating ecs-bridge bridge

sparrc:
The ecs-bridge bridge currently is created without setting a static MAC address. This means that the bridge MAC address inherits the 'lowest' MAC address of all of it's connected veth interfaces.

What this means for ECS tasks using awsvpc networking mode is that the MAC address of ecs-bridge can be changing as tasks are killed an added to an instance (on the ec2 launch type).

This presents a race condition where some tasks might have an arpcache entry pointing to the old MAC address when a task is killed and the ecs-bridge MAC changes. In rare cases this can lead to brief periods of lost network connectivity within the local network on the host (such as requests to the ECS Task Metadata Endpoint or the ECS credentials endpoint).

By setting a static MAC address at creation we disable this behavior of inheriting the 'lowest' MAC address. This means that the ecs-bridge MAC stays constant through the entire lifetime of the ECS container instance.

### Testing

Tested on a container instance using awsvpc network mode. Verified that MAC address is set and stays constant through many task starts/stops.

Ran ECS agent functional testing suite.

bugfix: Set static MAC address when creating ecs-bridge bridge

### Licensing

### DIFF
--- a/plugins/ecs-bridge/engine/engine.go
+++ b/plugins/ecs-bridge/engine/engine.go
@@ -14,9 +14,12 @@
 package engine
 
 import (
+	"crypto/rand"
 	"net"
 	"strings"
 	"syscall"
+
+	log "github.com/cihub/seelog"
 
 	"github.com/aws/amazon-ecs-cni-plugins/pkg/cniipamwrapper"
 	"github.com/aws/amazon-ecs-cni-plugins/pkg/cniipwrapper"
@@ -27,7 +30,7 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-const  (
+const (
 	// zeroLengthIPString is what we expect net.IP.String() to return if the
 	// ip has length 0. We use this to determing if an IP is empty.
 	// Refer https://golang.org/pkg/net/#IP.String
@@ -127,6 +130,22 @@ func (engine *engine) lookupBridge(bridgeName string) (*netlink.Bridge, error) {
 	return bridge, nil
 }
 
+// generateMACAddress generates a random locally-administrated MAC address.
+func (engine *engine) generateMACAddress() (net.HardwareAddr, error) {
+	buf := make([]byte, 6)
+	var mac net.HardwareAddr
+
+	_, err := rand.Read(buf)
+	if err != nil {
+		return mac, err
+	}
+
+	// Set locally administered addresses bit and reset multicast bit
+	buf[0] = (buf[0] | 0x02) & 0xfe
+	mac = append(mac, buf[0], buf[1], buf[2], buf[3], buf[4], buf[5])
+	return mac, nil
+}
+
 // createBridge creates a bridge interface
 func (engine *engine) createBridge(bridgeName string, mtu int) error {
 	bridgeLinkAttributes := netlink.NewLinkAttrs()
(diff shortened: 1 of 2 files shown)


plugins/ecs-bridge/engine/engine.go
```diff
@@ -143,6 +161,19 @@ func (engine *engine) createBridge(bridgeName string, mtu int) error {
 			"bridge create: unable to add bridge interface %s", bridgeName)
 	}
 
+	mac, err := generateMACAddress()
+	if err != nil {
+		return errors.Wrapf(err,
+			"bridge create: unable to generate mac addr %s", err)
+	}
+
+	log.Infof("Setting ecs-bridge hardware addr (MAC) %v %v", bridge, mac)
```
ofiliz, line 170: Looking at command.go and engine.go, it seems the logging is done by command.go before calling the implementation of engine in engine.go. We may want to stick with that convention and remove the log statement here.

plugins/ecs-bridge/engine/engine.go
```diff
@@ -14,10 +14,13 @@
 package engine
```
ofiliz, line 14: We should add an integration test which creates a bridge using this engine interface, attaches a dummy interface with a lower MAC address than the bridge MAC address. and confirms that the bridge MAC address does not change. The current mock tests won't work (since we need a real interface). However given the urgency of this fix, I don't want us to delay this PR. I can also add such a test as part of our unified networking library program.

ofiliz (changes requested):

plugins/ecs-bridge/engine/engine.go
sparrc: I would prefer to keep this log statement. Is there any drawback to logging this? Historically we've generally had issues with these plugins under-logging and not telling us what they are doing, rather than the opposite.

sparrc: there's virtually no risk of this spamming the logfile, as this function is generally only called once per instance.

ofiliz: We still log the error if the operation fails, correct?

plugins/ecs-bridge/engine/engine.go
```diff
@@ -127,6 +130,21 @@ func (engine *engine) lookupBridge(bridgeName string) (*netlink.Bridge, error) {
...
+	var mac net.HardwareAddr
+
+	_, err := rand.Read(buf)
+	if err != nil {
+		return mac, err
+	}
+
+	// Set locally administered addresses bit and reset multicast bit
+	buf[0] = (buf[0] | 0x02) & 0xfe
+	mac = append(mac, buf[0], buf[1], buf[2], buf[3], buf[4], buf[5])
```
ubhattacharjya, line 144: QQ: Can this MAC address ever collide with any of the other interfaces? Do we need to have a check here?

sparrc: I'm not sure how practical checking this would be at this point. It is very unlikely, I did find this on the internet, for what it's worth (keep in mind this is also a local bridge which maxes out on the order of 1000 interfaces):

MAC addresses are 48 bits long, so there are 281,474,976,710,656 possible MAC addresses. In a network consisting of 1,000,000 (one million) devices, the chance of random MAC address collision is 0,0000000036% (about one in 280,000,000). That's 1000 times less probable than being struck by a lightning.

plugins/ecs-bridge/engine/engine.go
sparrc: yes, but issues where there's an explicit failure are always the easiest to debug :)

ofiliz: I still don't think this log statement provides any additional value, but don't have a strong opinion. Let's go with your approach.

ofiliz (approved):
Thank you!

ubhattacharjya (approved):

PR #120, opened 2026-02-24
Bridge plugin updates to define a route from replica task to daemon t…

aviral92:
…ask, via the bridge.

<!--
Please make sure you've read and understood our contributing guidelines;
https://github.com/aws/amazon-ecs-init/blob/master/CONTRIBUTING.md

Please provide the following information:
-->

### Summary
<!-- What does this pull request do? -->
There are changes in this PR for the bridge, that allow to setup a route from a replica task to the daemon task, via the bridge. Corresponding ecs-agent changes - https://github.com/aws/amazon-ecs-agent/pull/4881. 

### Implementation details
- Bridge /22 Mask for LinkLocal IPs. Link-local IPs (169.254.x.x) need /22 mask on bridge to cover the entire ECS subnet (169.254.172.0/22), allowing proper routing.
- On-Link Gateway Routes - With /32 mask, the gateway (169.254.172.1) is not in the same subnet as the daemon container IP (169.254.172.2/32). An on-link route tells the kernel that this gateway is directly reachable even though it's not in my subnet.
- Connected Subnet Routes - Tells the kernel "to reach the entire 169.254.172.0/22 subnet, use this interface with this source IP". This enables proper routing for the connected network.
- Default Route Deletion Logic would delete the on-link gateway routes we just added, hence had to add logic to preserve the routes we added.

## Bridge Plugin: Preserve Routes for AWSVPC-Only Tasks

## Problem
AWSVPC-only tasks (without daemon) were incorrectly getting the /22 connected subnet route (169.254.172.0/22 dev eth0 scope link), which is only needed for daemon-bridge communication. This route was being added unconditionally based on the IP mask from IPAM, ignoring the connectedSubnetMaskSizeIPv4 
parameter.

## Solution
Modified the CNI bridge plugin to conditionally add the connected subnet route based on the connectedSubnetMaskSizeIPv4 parameter:
- When connectedSubnetMaskSizeIPv4 = 0 (default, awsvpc-only): Skip adding the /22 route
- When connectedSubnetMaskSizeIPv4 = 22 (daemon exists): Add the /22 route for replica-daemon communication

Additionally fixed route cleanup logic to preserve /32 and /128 host routes (gateway routes) during cleanup.

## Changes

### Core Logic (configureveth.go)
1. Lines 138-156: Skip adding connected subnet route when mask size is 0
2. Lines 118-133: Exclude /32 and /128 host routes from deletion during cleanup
3. Lines 180-210: Add connected subnet route only when mask size is non-zero

### API Changes (types.go, commands.go, engine.go)
- Added connectedSubnetMaskSizeIPv4 and connectedSubnetMaskSizeIPv6 parameters to ConfigureContainerVethInterface
- Parameters flow from bridge configuration through to veth configuration context

### Tests
- Added 3 new unit tests covering the conditional route logic and host route preservation
- Updated existing tests to use appropriate mask size parameters (22 for IPv4, 64 for IPv6)
- All tests passing

## Validation
Tested on live instances:
- **AWSVPC-only** (i-07125862b3f92aa0e): ✅ No /22 route, gateway route present, TMDS accessible
- **AWSVPC+daemon** (i-08adc00b03c046850, i-0d8875279f5fb2087): ✅ Has /22 route, daemon reachable, TMDS accessible

## Expected Routes

### AWSVPC-only task (single namespace)
default via 10.194.20.1 dev eth1               # Default route via task ENI
10.194.20.0/24 dev eth1 scope link             # Task ENI subnet                                                                                                                                                                                                                                                    
blackhole 169.254.169.254                      # Block IMDS                                                                                                                                                                                                                                                         
169.254.170.2 via 169.254.172.1 dev eth0      # TMDS route via gateway                                                                                                                                                                                                                                              
169.254.172.1 dev eth0 scope link              # Gateway route                                                                                                                                                                                                                                                      
### No /22 route  above.                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    

### AWSVPC+daemon - Replica task namespace
default via 10.194.20.1 dev eth1               # Default route via task ENI
10.194.20.0/24 dev eth1 scope link             # Task ENI subnet                                                                                                                                                                                                                                                    
blackhole 169.254.169.254                      # Block IMDS                                                                                                                                                                                                                                                         
169.254.170.2 via 169.254.172.1 dev eth0      # TMDS route via gateway                                                                                                                                                                                                                                              
169.254.172.0/22 dev eth0 scope link           # Connected subnet for daemon communication                                                                                                                                                                                                                          
169.254.172.1 dev eth0 scope link              # Gateway route                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                    

### AWSVPC+daemon - Daemon namespace (host-daemon)
default via 169.254.172.1 dev eth0             # Default via gateway
blackhole 169.254.169.254                        # Block IMDS  
169.254.170.2 via 169.254.172.1 dev eth0      # TMDS route via gateway                                                                                                                                                                                                                                              
169.254.172.1 dev eth0 scope link              # Gateway route                                                                                                                                                                                                                                                      
169.254.172.2 dev eth0 scope link              # Daemon static IP



-----

## Summary for block IMDS changes.

Add IMDS (Instance Metadata Service) blocking support to the ECS bridge CNI plugin, allowing callers to block access to both IPv4 and IPv6 IMDS endpoints via configuration.

## Changes

Bridge CNI Plugin (amazon-ecs-cni-plugins)

### Core Implementation
- **Added BlockInstanceMetadata() method** to the Engine interface that blocks IMDS access within network namespaces
- **Blocks both IPv4 and IPv6 endpoints**: 169.254.169.254/32 and fd00:ec2::254/128
- **Uses blackhole routes** for blocking (consistent with vpc-branch-eni plugin approach)
- **Resilient to unsupported IP versions**: Continues if one IP version fails (e.g., IPv6 on IPv4-only systems)

### Configuration
- **Added BlockIMDS field** to BridgeConfig struct (maps to blockInstanceMetadata in JSON)
- **Integrated into command flow**: Bridge plugin checks config and blocks IMDS when enabled

### Testing


Added UTs and tested with custom AMI.


idge plugin updates to define a route from replica task to daemon task, via the bridge.

### Licensing


plugins/ecs-bridge/commands/commands.go
```diff
@@ -133,17 +133,21 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 		result.IPs[i].Interface = 2
 	}
 
-	detailLogInfo("Configuring container's interface", args, conf, hostVethName)
-	err = engine.ConfigureContainerVethInterface(args.Netns, result, args.IfName)
+	// Configure bridge first so gateway IP exists before adding container routes
```
JoseVillalta, line 136: nit, non-blocking: Comments should end with a period. You can set a rule in Kiro to do this for you.

JoseVillalta (approved):

plugins/ecs-bridge/commands/commands.go
```diff
@@ -133,17 +133,21 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
...
+	detailLogInfo("Configuring bridge", args, conf, hostVethName)
```
aaithal, line 137: this seems redundant. you're logging the same thing 3-4 lines down.

plugins/ecs-bridge/engine/configureveth.go
```diff
@@ -113,25 +131,41 @@ func (configContext *configureVethContext) run(hostNS ns.NetNS) error {
...
+		}
+		
+		// Check if this is a default route
+		isDefaultRoute := route.Dst == nil || 
+			route.Dst.String() == "0.0.0.0/0" || 
+			route.Dst.String() == "::/0"
+		
+		// Check if this is a connected subnet route that should be deleted (awsvpc-only mode)
+		// Exclude /32 (IPv4) and /128 (IPv6) routes as these are host routes (like gateway routes)
+		isConnectedSubnetToDelete := false
```
aaithal, line 157: Can you help me understand what's happening in this block? Ideally, we won't need to care about awsvpc etc in this plugin. So, I want to understand what's the special casing here. 

I am essentially wondering why the plugin needs to make these decisions. Why can't the caller pass in a config, which tells the plugin exactly what to do. Can we move this intelligence up the stack a little bit?

plugins/ecs-bridge/types/types.go
```diff
@@ -24,12 +24,20 @@ import (
...
+
 // NetConf defines the parameters required to configure a bridge and to
 // attach the same to the contaner's namespace
 type NetConf struct {
 	types.NetConf
-	BridgeName string `json:"bridge"`
-	MTU        int    `json:"mtu"`
+	BridgeName string      `json:"bridge"`
+	MTU        int         `json:"mtu"`
+	IPAM       *IPAMConfig `json:"ipam,omitempty"`
```
aaithal, line 40: Why is bridge netconf accepting IPAM conf? Shouldn't that be something that the IPAM plugin accepts?

plugins/ecs-bridge/commands/commands.go
aviral92: sure, will fix.

plugins/ecs-bridge/commands/commands.go
aviral92: This was mostly for debugging, will remove this redundant comment.

plugins/ecs-bridge/types/types.go
aviral92: This is existing code structure in the bridge plugin - the BridgeNetConf has always contained 
the IPAM configuration. The bridge plugin is responsible for:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 - Creating the bridge interface ,                                                                                                                                                                                                                                                                                Then, Calling the IPAM plugin to allocate IPs ,                                                                                                                                                                                                                                                                         And then, Configuring the veth pair with those IPs                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                    
The IPAM field in BridgeNetConf is used to pass configuration to the IPAM plugin when the bridge                                                                                                                                                                                                                    
plugin invokes it. Basically the bridge plugin orchestrates other plugins (IPAM).                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
My changes only added the connectedSubnetMaskSizeIPv4/IPv6 fields to the existing IPAMConfig                                                                                                                                                                                                                        
structure - I didn't change the overall architecture of how bridge and IPAM interact.  Its basically an override to the existing struct which was anyways getting passed.

plugins/ecs-bridge/engine/configureveth.go
aviral92: Good question.
The intelligence is actually in the caller (ecs-agent) - it passes connectedSubnetMaskSizeIPv4                                                                                                                                                                                                                      
parameter to tell the plugin what to do:                                                                                                                                                                                                                                                                            
- connectedSubnetMaskSizeIPv4 = 0 → Don't add connected subnet route (awsvpc-only)                                                                                                                                                                                                                                  
- connectedSubnetMaskSizeIPv4 = 22 → Add /22 connected subnet route (daemon exists)                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
The block below is checking this parameter to decide whether to                                                                                                                                                                                                                       
delete existing connected subnet routes during cleanup. The logic is:                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                    
1. If maskSize = 0 (awsvpc-only), we need to DELETE any existing /22 routes                                                                                                                                                                                                                                         
2. If maskSize = 22 (daemon mode), we PRESERVE the /22 routes                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                    
The `awsvpc-only mode` comment is just for clarity - the actual decision is based purely on the                                                                                                                                                                                                                     
maskSize parameter from the caller. The plugin doesn't independently determine the mode.  Happy to update the comments so that its clear for readers in future.

plugins/ecs-bridge/types/types.go
aaithal: Ack, thanks.

plugins/ecs-bridge/engine/configureveth.go
aaithal: ok, that makes sense. Let's make the comments a bit more clear then. But, not blocking the PR on that.

aaithal (approved):

plugins/ecs-bridge/commands/commands.go
```diff
@@ -148,6 +148,15 @@ func add(args *skel.CmdArgs, engine engine.Engine) error {
 		return err
 	}
 
+	// Block IMDS if configured
+	if conf.BlockIMDS {
```
JoseVillalta: should we check that conf is not nil? what happens if we call it and it's not set?

JoseVillalta (approved):
Please update documentation if these changes are not backwards compatible.

aviral92:
> Please update documentation if these changes are not backwards compatible.

These changes are backwards compatible:
                                                                                                                                                                                                                                                                                                                    
1. **BlockIMDS field**: Defaults to false if not set, so existing configurations without                                                                                                                                                                                                                            
   this field will continue to work as before (no IMDS blocking).                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                    
2. **connectedSubnetMaskSizeIPv4/IPv6 parameters**: Default to 0, which preserves the                                                                                                                                                                                                                               
   existing behavior for awsvpc-only tasks (no /22 route added).                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
The new functionality is opt-in via configuration, so no breaking changes for existing                                                                                                                                                                                                                              
deployments.

plugins/ecs-bridge/commands/commands.go
aviral92: Good catch, let me update code to add the nil check, that way we wont panic.

plugins/ecs-bridge/engine/configureveth.go
aviral92: Done, updated.

plugins/ecs-bridge/commands/commands.go
aviral92: Done, removed.

PR #124, opened 2026-05-14
fix(ecs-bridge): Make BlockInstanceMetadata idempotent for EEXIST

JoseVillalta:
### Summary

Fix daemon task launch failures caused by `BlockInstanceMetadata` failing with "file exists" when IMDS blackhole routes already exist in a persistent daemon network namespace.

### Implementation details

The daemon network namespace persists for the instance lifetime. When `StopDaemonNetNS()` tears down a daemon task, it removes the veth pair via CNI DEL but the IMDS blackhole routes remain (they have no device association — just `RTN_BLACKHOLE` with a `Dst`, no `LinkIndex`). On the next daemon task start, the agent invokes the ecs-bridge CNI plugin with `blockInstanceMetadata: true`, and `BlockInstanceMetadata()` calls `RouteAdd()` for a route that already exists, returning `EEXIST`.

The fix adds an `os.IsExist(err)` check in `blockIMDSContext.run()` (ecs-bridge plugin) and in the `setupNamespaceClosure` IMDS blocking loop (eni plugin). When `RouteAdd` returns `EEXIST`, the error is treated as success since the desired state (IMDS blocked) is already achieved, and execution continues to the next endpoint.

Files changed:
- `plugins/ecs-bridge/engine/engine.go` — Handle `EEXIST` in `blockIMDSContext.run()`
- `plugins/eni/engine/nsclosure.go` — Same fix in the eni plugin's IMDS blocking loop
- `plugins/ecs-bridge/engine/engine_test.go` — Unit tests for the idempotency fix
- `plugins/ecs-bridge/e2eTests/e2e_test.go` — E2E test validating bridge re-creation after deletion

### Testing

- Unit tests: 4 new test cases in `TestBlockIMDSContextRun` covering routes-don't-exist (existing behavior), routes-already-exist (EEXIST fix), non-EEXIST errors still propagate, and mixed EEXIST/real-error scenarios. All pass.
- E2E test: `TestAddBridgeAfterDeletion` validates the full ADD → DEL → delete bridge → ADD cycle succeeds. Passes on remote Linux host.
- Remote build (`make plugins`) passes cleanly.

New tests cover the changes: yes

Bug - Make BlockInstanceMetadata idempotent by handling EEXIST when IMDS blackhole routes already exist in persistent daemon network namespaces

### Licensing

### DIFF
--- a/plugins/eni/engine/nsclosure.go
+++ b/plugins/eni/engine/nsclosure.go
@@ -16,6 +16,7 @@ package engine
 import (
 	"fmt"
 	"net"
+	"os"
 	"syscall"
 
 	"github.com/aws/amazon-ecs-cni-plugins/pkg/netlinkwrapper"
@@ -27,7 +28,7 @@ import (
 var (
 	// InstanceMetadataEndpoints is the list of EC2 instance metadata endpoints.
 	InstanceMetadataEndpoints = []string{"169.254.169.254/32", "fd00:ec2::254/128"}
-	linkWithMACNotFoundError = errors.New("engine: device with mac address not found")
+	linkWithMACNotFoundError  = errors.New("engine: device with mac address not found")
 )
 
 // setupNamespaceClosureContext wraps the parameters and the method to configure the container's namespace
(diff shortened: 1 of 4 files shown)


JoseVillalta:
Do not merge yet, will do an end to end test where we repro the issue and then validate the fix

plugins/eni/engine/nsclosure.go
```diff
@@ -147,19 +148,24 @@ func (closureContext *setupNamespaceClosureContext) run(_ ns.NetNS) error {
...
+			if err = closureContext.netLink.RouteAdd(&netlink.Route{
+				Dst:  imdsNetwork,
+				Type: syscall.RTN_BLACKHOLE,
+			}); err != nil {
+				// If the route already exists, treat it as success since the
+				// desired state (IMDS blocked) is already achieved.
+				if os.IsExist(err) {
+					continue
+				}
+				return errors.Wrapf(err, "setupNamespaceClosure engine: unable to add route to block instance metadata")
```
xxx0624, line 166: non-blocking - will it be helpful to include the route details in the error?

xxx0624 (approved):

aviral92 (approved):
LGTM. Looks like this is the only place where the route exists check did not exist. 

Checked - The gateway route (/32 host route to gateway) and the connected subnet route (/22 subnet route) in configureveth.go already use the !os.IsExist(err) pattern. The ConfigureBridge function also has an addressExists check before adding the bridge IP address, so that's covered too.

Please feel free to think more on this, the above is my analysis of this change. And I see you are doing more testing, please share results once all looks well.

amogh09 (approved):

JoseVillalta:
Tested the changes with end to end test with a special AMI that can create daemon namespaces from regular RunTask API requests, validated that this change fixes the issue, was able to reproduce the bug without the fix.

